Issue #1, opened 2014-07-08
Account revaluation always takes company_id from user [7.0]


kenvac:
Hi Team,

I was recently testing account_revaluation module for multi company having multi currency and found a bug.

The details are as follows:
- Account periods are always searched from user's company instead of selected journal. 
  "company = user_obj.browse(cr, uid, uid).company_id"
  accounting to me this should be
  "company = form.journal_id.company_id or user_obj.browse(cr, uid, uid).company_id"

It should search period for company in journal rather than logged in user.

PR #2, opened 2014-07-09
[Fix] user company_id instead of journal company_id

kenvac:
Fix for issue #1

### DIFF
--- a/account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py
+++ b/account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py
@@ -162,7 +162,7 @@ class WizardCurrencyrevaluation(orm.TransientModel):
         ctx_rate['date'] = form.revaluation_date
         ctx_rate['currency_rate_type_id'] = type_id
         user_obj = self.pool.get('res.users')
-        cp_currency_id = user_obj.browse(cr, uid, uid, context=context).company_id.currency_id.id
+        cp_currency_id = form.journal_id.company_id.currency_id.id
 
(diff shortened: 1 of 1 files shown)


bwrsandman:
Is this for the 8.0 branch?

pedrobaeza:
:+1:

kenvac:
Its version 7.0. But I'm sure the bug exist for version 8.0 as well.

bwrsandman:
@kenvac just so you know this PR is proposed to the master branch which is not 7.0

I suggest you make another PR for 7.0.

You can do so with the same branch:  kenvac:master but select  OCA:7.0 as the base branch

pedrobaeza:
@bwrsandman, you should consider both branches, not only 7.0, because bug will be also there.

bwrsandman:
Porting process of 8.0 for these modules has not started: last commit on master is in January

If there is a merge here, it is likely that commit will be lost or squashed when that process happens.

pedrobaeza:
Yeah, but if we don't include this patch, v8.0 will have a regression against the v7 branch. The other alternative is to rebase from 7.0 when integrated.

bwrsandman:
Rebase, reset or merge, we could even create 8.0 from 7.0.
Since the nomenclature of version branches for 8.0 isn't clear, I wouldn't trust commiting here unless it's oviously the proper place, and in this case, it was a wrong base branch.

Note that if you checkout odoo master, it will identify as Odoo 9 Alpha 1.

pedrobaeza:
OK, let's wait until we have it clear.

Regards.

guewen:
The 8.0 branch is there so this PR should be submitted again against the 8.0 branch.

PR #22, opened 2015-05-05
[ADD] Add account_move_check_period module

ehdem:
This module adds contraint when creating an entry
with a closed period

### DIFF
--- a/account_move_check_period/__init__.py
+++ b/account_move_check_period/__init__.py
@@ -1,0 +1,25 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2015 - Present Savoir-faire Linux
+#    (<http://www.savoirfairelinux.com>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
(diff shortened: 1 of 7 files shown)


sbidoul:
Hello @ehdem 

Can you elaborate a bit the purpose of the module? I would say that if Odoo lets the user enter an entry in a closed period it is a bug that should be resolved with an Odoo and/or OCB PR.

Also note that this repository is dedicated to end-of-period cutoff so it is probably not the right place for such a module.

As a side note, you may be interested in this module: https://github.com/OCA/account-financial-tools/tree/8.0/account_journal_period_close (also available in 7.0).

ehdem:
@sbidoul : For example, you can create manually a journal item with a closed period.
Go to Accounting/Journal entries/Journal items; in the form you can list of closed and opened period.

sbidoul:
@ehdem at least on 8.0 you can't create entries in closed period from the journal items view.
Closed periods are in the list because this view is also used for searching so that makes sense.

So if 7.0 allows it, it is clearly a bug and I think it should be treated with an upstream PR instead of a module (I see no reason why Odoo would not accept a PR fixing that on 7.0).

sbidoul:
At this stage, I'm :-1: for this one as I believe this should be fixed upstream and/or in OCB.

sbidoul:
No feedback for a long period. Closing.

PR #21, opened 2015-04-24
8.0 account_invoice_accrual

sbidoul:
This new module lets the user generate accrual move for supplier (or customer) invoices with one click on the draft invoice form.

When goods or services are received, the draft invoice is generated using standard odoo processes. If at the end of the period, the supplier invoice has not been received, the user can click "Accrue" on the invoice to generate an accrual move in the period being closed. Next period when the supplier invoice arrives, the accrual move is automatically reversed when the invoice is valided.

It is based on account_cutoff_accrual_base account mapping framework.

This PR is still WIP while we fix the last issues being discovered during testing.

This PR depends on #8

PR #24, opened 2015-07-01
[ADD][account_cutoff_accrual_base] Add a default journal configuration for each type of cutoff

adrienpeiffer:
Add a default journal configuration for each type of cutoff.

The default cutoff journal is kept for backward compatibility and as a fallback if no default journal is defined for a specific type of cutoff.

### DIFF
--- a/account_cutoff_accrual_base/account_cutoff.py
+++ b/account_cutoff_accrual_base/account_cutoff.py
@@ -48,9 +48,14 @@ class account_cutoff(orm.Model):
             ._get_default_journal(cr, uid, context=context)
         cur_user = self.pool['res.users'].browse(cr, uid, uid, context=context)
         cutoff_type = context.get('type', False)
-        if cutoff_type in ['accrued_expense', 'accrued_revenue']:
-            journal_id = cur_user.company_id.\
-                default_accrual_journal_id.id or False
+        default_journal_id = cur_user.company_id\
+            .default_cutoff_journal_id.id or False
+        if cutoff_type == 'accrued_expense':
+            journal_id = cur_user.company_id\
+                .default_accrual_expense_journal_id.id or default_journal_id
+        elif cutoff_type == 'accrued_revenue':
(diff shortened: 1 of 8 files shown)


sbidoul:
I have doubts about this one. Because if we push the logic we may want different default also for prepaid expenses and prepaid revenue and accrued expense and accrued revenue... and also based on the type of accrual you do (invoice, picking, timesheet) so it's potentially a never ending story.

Would it not be better to set the default in the wizard field directly?

adrienpeiffer:
@sbidoul, as discussed offline, I added one default journal by cutoff type

account_cutoff_accrual_base/account_cutoff.py
```diff
@@ -43,6 +43,21 @@ def _inherit_default_cutoff_account_id(self, cr, uid, context=None):
...
+        cutoff_type = context.get('type', False)
+        default_journal_id = cur_user.company_id\
+            .default_cutoff_journal_id.id or False
+        if cutoff_type == 'accrued_expense':
+            journal_id = cur_user.company_id\
+                .default_accrual_expense_journal_id.id or default_journal_id
+        elif cutoff_type == 'accrued_revenue':
+            journal_id = cur_user.company_id\
+                .default_accrual_revenue_journal_id.id or default_journal_id
+        return journal_id
```
sbidoul, line 61: perhaps I'm nitpicking but I dont like the place the backslash, it would be easier to read with the backslash after `or`

sbidoul:
I updated the PR title to better reflect what it does.

sbidoul:
@adrienpeiffer can you also reflect this change in #21 ?

account_cutoff_accrual_base/account_cutoff.py
adrienpeiffer, line 61: with that, the length of the line is greater than 79

sbidoul:
:+1:

pedrobaeza:
:+1:

PR #26, opened 2015-07-06
[IMP] account_cutoff_base: cutoff moves always need to be reversed

sbidoul:
So we set the to_be_reversed flag so it is easy to find the moves
that have to be reversed and not forget them.

### DIFF
--- a/account_cutoff_base/__openerp__.py
+++ b/account_cutoff_base/__openerp__.py
@@ -42,7 +42,10 @@ for any help or question about this module.
     """,
     'author': "Akretion,Odoo Community Association (OCA)",
     'website': 'http://www.akretion.com',
-    'depends': ['account_accountant'],
+    'depends': [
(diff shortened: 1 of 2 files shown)


adrienpeiffer:
:+1:

pedrobaeza:
Don't you think we should provide this operation automatically when closing fiscal year?

sbidoul:
@pedrobaeza Accountants I have met like to do it manually. Reversals are done also at the end of period, not only end of fiscal year.

Perhaps automating reversals could be done in an another generic module. I've not met that requirement yet.

pedrobaeza:
Thanks for your input about the topic. I'm not sure how to proceed, because Spain have an specific fiscal year closing... I'll tell you when I think about it.

About this change, although it introduces a new dependency, I think it's tolerable.

pedrobaeza:
:+1:

PR #33, opened 2015-11-12
[IMP] account_cutoff_prepaid: Add fields explicitely on move line form+tree view inside account.move

pedrobaeza:


pedrobaeza:
cc @antespi @janetrp

rafaelbn:
Functional review , it looks good but runbot fails

pedrobaeza:
@sbidoul, @adrienpeiffer, can you review?

sbidoul:
Code looks good. I'm a bit afraid this tree view will become too crowded. I'll test on runbot when ready.

pedrobaeza:
Yeah, I know, but it's a must because you can't access form view from the account.move view.

rafaelbn:
Tested in runbot :+1:

rafaelbn:
ping @antespi

sbidoul:
@pedrobaeza @rafaelbn I'm curious, do you have users willing to enter start/end dates on miscellaneous operations? Ours are happy with having that on invoices only.

pedrobaeza:
Yes, because there are some manual entries that they need to enter/modify.

sbidoul:
@pedrobaeza Yeah but I mean which type of entries? I my experience all entries that are subject to cutoff are also invoices. So I'm curious.

pedrobaeza:
I didn't remember well, but when seeing with the customer, we ended up with this need. @rafaelbn, do you remember?

pedrobaeza:
@sbidoul, I have remembered the use case: some insurances payments are registered via bank statement, which generates an account entry. This doesn't have any invoices, but goes directly against an expense account. This expense has to be cutoff, and the only way to put the dates we have is through journal entry screen.

sbidoul:
Ok, thanks a lot for the explanation. Let's see if user complain about the crowded screen or not.

PR #34, opened 2015-11-12
add oca_dependencies.txt

sbidoul:


pedrobaeza:
Please remove the travis lines: https://github.com/OCA/account-closing/blob/9.0/.travis.yml#L32

sbidoul:
@pedrobaeza done

pedrobaeza:
Why adding reporting-engine?

sbidoul:
Because report_xls is a dependency of account_asset_management.

pedrobaeza:
But account_asset_management is not installed...

sbidoul:
But it is imported. I've seen an `import report_xls`  fail earlier today in this project.

pedrobaeza:
Indeed the calls are not protected according current guidelines, but we can stay as is for now.

I fast-track it to get green branches.

PR #35, opened 2015-12-21
Multi currency revaluation with rate type

levkar:
Adds a currency rate type module and multi currency revaluation with rate type feature


levkar:
Few comments for the reviewer:

1) I'm not sure if this is the right place for currency_rate_type addon.
2) I used mostly smile's similar addon for currency_rate_type with few fixes. I added them as reference into code. I really don't know if that will be enough.
3) Rate type is added to individual accounts instead of main form since different accounts can be revaluated with different rates (this is how it works in my country).

levkar:
Is there anyone who can review it?

currency_rate_type/README.rst
```diff
@@ -0,0 +1,80 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================
+Currency Rate Type
+==================
+
+This module adds currency rate types to currency.
```
gurneyalex, line 9: could you give an example of rate types here?

gurneyalex (commented):
I'm indeed not sure the currency rate type belongs in account closing, but I'm not a specialist. Did you try discussing this on the accounting mailing list of the OCA?

levkar:
@gurneyalex : I agree with you. We can relocate it in another repo.

Rate type exists in v7 and removed in v8. So in a way we can consider this as a forward port. Rate type is selectable in v7, here the main difference is that we put the rate type onto account. You specify rate type and that account is revaluated with the selected rate type. It has been discussed on the mailing list.

A real world example is like this:

Central bank in Turkey declares "forex buying" and "banknote buying" rates seperately and when revaluating cash accounts "banknote buying" rate is used and other rates are revaluated with "forex buying" rate. So we select "banknote buying" rate on types of cash and "forex buying" on others.

And here's a simpler example:

A rate type so called "Revaluation rate" is created and it's value for revaluation date is entered manually. It is selected on account to be revaluated and it uses that rate while revaluating.

Actually we have a serious problem with revaluation I need to ask on the mailing list. It doesn't work after year close since they are reconciled on year close and revaluation doesn't take reconciled moves into account...

feketemihai:
@levkar do you still need this on v8, i agree that in some cases rate type can be useful, but we need a more generic way to do it without overwriting base methods of getting rates.

levkar:
Not any more. I am closing this.

PR #36, opened 2016-06-29
FIX error in view account_invoice_view

drl9319:
I found a error when I installes this module becouse in my system have check four settings in Configuration -> Accounting: 
1. Analytic accounting for sales
2. Use multiple analytic accounts on sales
3. Analytic accounting for purchases
4. Use multiple analytic accounts on orders

when this four settings are check, Odoo will install sale_analytic_plans and purchase_analytic_plans addons, this addons delete account_analytic_id field from invoice form. For this when this addon needs insert two fields after this have a problem, for fix this problem I think the best solutions is insert this fields before quantity field, the place of this two field it is the same and quantity field is never delete.

### DIFF
--- a/account_cutoff_prepaid/README.rst
+++ b/account_cutoff_prepaid/README.rst
@@ -25,6 +25,7 @@ Contributors
 
 * Alexis de Lattre <alexis@via.ecp.fr>
 * Stéphane Bidoul <stephane.bidoul@acsone.eu>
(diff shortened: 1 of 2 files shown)


gurneyalex (approved):
LGTM.

pedrobaeza (approved):
Damn modules that make a "replace" operation...

PR #37, opened 2016-07-24
9.0 Port account_cutoff_base and account_cutoff_prepaid

alexis-via:
Extract a new module account_invoice_start_end_dates from account_cutoff_prepaid

Warning: it is not a port of the OCA/account-closing/8.0 version of the module, but a port of my LP branch https://code.launchpad.net/~akretion-team/account-closing/80-forecast-prepaid which is still the branch I use for Odoo v8 (which explains why I wasn't active on OCA/account-closing so far). This is due to the fact that my "forecast" branches were not merged on LP before the transition to github.

So the next step is to merge the few enhancements that took place in the OCA version of the module.


alexis-via:
And of course, this PR is also a port to the new API.

alexis-via:
@adrienpeiffer I don't see the pull request you told me about...

adrienpeiffer (approved):
:+1: Code review and functional test

adrienpeiffer:
@chrisdec do you have a moment to test it on runbot ? :smile:

gurneyalex (approved):
code review, no test.

account_cutoff_base/models/account_cutoff.py
```diff
@@ -0,0 +1,359 @@
...
+            label = _('Accrued Revenue')
+        elif type == 'prepaid_revenue':
+            label = _('Prepaid Revenue')
+        elif type == 'prepaid_expense':
+            label = _('Prepaid Expense')
+        return label
+
+    @api.model
+    def _inherit_default_cutoff_account_id(self):
+        '''Function designed to be inherited by other cutoff modules'''
```
pedrobaeza: Docstrings should use double triple-quota (applicable to all)

account_cutoff_base/models/account_cutoff.py
```diff
@@ -0,0 +1,359 @@
...
+                'name': move_label,
+                'debit': amount < 0 and amount * -1 or 0,
+                'credit': amount >= 0 and amount or 0,
+            }
+            for k, v in zip(merge_keys, merge_values):
+                vals[k] = v
+            movelines_to_create.append((0, 0, vals))
+            amount_total += amount
+
+        # add contre-partie
```
pedrobaeza: counter-part

account_cutoff_base/security/account_cutoff_base_security.xml
```diff
@@ -1,11 +1,20 @@
...
-        <record id="account_cutoff_multi_company_rule" model="ir.rule">
-            <field name="name">Account Cutoff Multi Company</field>
-            <field name="model_id" ref="model_account_cutoff"/>
-            <field name="global" eval="True"/>
-            <field name="domain_force">['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]</field>
-        </record>
-    </data>
-</openerp>
+<odoo>
+<data noupdate="1">
```
pedrobaeza: Put directly `<odoo noupdate="1">`

account_cutoff_base/views/account_cutoff.xml
```diff
@@ -1,30 +1,28 @@
...
-    Copyright (C) 2013 Akretion (http://www.akretion.com/)
-    @author Alexis de Lattre <alexis.delattre@akretion.com>
-    The licence is in the file __openerp__.py
+  © 2013-2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 -->
 
-<openerp>
+<odoo>
 <data>
```
pedrobaeza: You can remove `<data>` tag

account_cutoff_base/views/account_cutoff.xml
```diff
@@ -35,15 +33,15 @@
...
+                        <field name="cutoff_date"/>
+                        <field name="total_cutoff_amount"/>
                         <field name="company_id" groups="base.group_multi_company" widget="selection" />
                         <field name="company_currency_id" invisible="1"/>
                     </group>
                     <group name="accounting-params">
-                        <field name="cutoff_journal_id"/>
-                        <field name="cutoff_account_id"/>
-                        <field name="move_label"/>
+                        <field name="cutoff_journal_id" required="1"/>
```
pedrobaeza, line 42: Shouldn't this be put (required flag) at field level?

account_cutoff_prepaid/i18n/account_cutoff_prepaid.pot
```diff
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
```
pedrobaeza: Remove POT file

account_invoice_start_end_dates/demo/product_demo.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  © 2013-2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+-->
+
+<odoo>
+<data>
```
pedrobaeza: Remove `<data>` tag

account_invoice_start_end_dates/models/account_invoice.py
```diff
@@ -0,0 +1,91 @@
...
+            # block automatic invoice generation/import. So we do the check
+            # upon validation of the invoice (see below the function
+            # action_move_create)
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    def inv_line_characteristic_hashcode(self, invoice_line):
+        '''Add start and end dates to hashcode used when the option "Group
```
pedrobaeza: Double quote

pedrobaeza (changes requested):
Several change across full code

account_cutoff_base/models/account_cutoff.py
alexis-via: If docstring should really use double triple-quota, then why flake8/pylint says nothing about it ?

chrisdec:
Hello, 

Functional test : 

Just one note : is it possible to show the message "Don't forget to Re-Generate Lines after entering or leaving forecast mode." only when user actives or disables the checkbox "forecast"? It's quite confusing to have this message when you open a new prepaid expense or a new prepaid revenue.

Except that, 👍

account_cutoff_base/models/account_cutoff.py
pedrobaeza: Yes, good point. We need to add that check. But you can see here that what I say is correct: https://www.python.org/dev/peps/pep-0257/#id15 (`For consistency, always use """triple double quotes""" around docstrings. Use r"""raw triple double quotes""" if you use any backslashes in your docstrings. For Unicode docstrings, use u"""Unicode triple-quoted strings""" .`)

account_cutoff_base/views/account_cutoff.xml
alexis-via: It is set at view level because it is inherited in the module account_cutoff_prepaid and replaced by attrs = {'required': [('forecast', '=', False)]}

alexis-via:
@chrisdec Yes, we need to find a better way. I guess a better way would be to use a button to enter/leave forecast mode (like the "Active" button on partners for example) and delete all lines when the user clicks on this button. I'll try to find time to do that...

account_cutoff_base/views/account_cutoff.xml
pedrobaeza: OK, thanks for the explanation. A comment in the XML would save this question for future readers of the code.

pedrobaeza (approved):

PR #65, opened 2018-01-19
[8.0] account_cutoff_* : allow use by Accountant

alexis-via:
allow use by Accountant, not just by Financial Manager.

The Cutoff objects are not "configuration" items, but regular accounting use, so it should be accessible by default by the Accountant.

If you agree on this, I'll up-port this PR to newer versions.


alexis-via:
@sbidoul Do you agree on this ? If yes, I'll merge it.

fclementic2c:
:+1:

sbidoul:
:+1:

PR #66, opened 2018-01-22
[10.0][IMP] Traduccion al español

pepetreshere:


pedrobaeza:
Why don't you use Transifex?

pepetreshere:
Hi Pedro!... didn't knew about transifex... I will google it, but if by chance you have at hand some tutorial please share with me.

pedrobaeza:
Please see https://odoo-community.org/page/translate

pepetreshere:
Hi Pedro, I have been accepten in Transifex OCA team por Spanish EC, but not por Spanish ES. I would like to be accepted for Spanish ES also.

The reason is we use the Spanish ES tranlation for many projects in Ecuador as the terms are very similar and for historical reasons in our older deployments of Odoo and OpenERP.

pedrobaeza:
Yeah, it was me, because main Spanish language is for Spain, and it has been previously changed by other regionalisms, so it's better to directly translate in your regional variant.

pepetreshere:
:) oks Pedro. I will keep this translation in my branch meanwhile (as for this project we are based on ES terms + some modifications in a custom module, so not possible for now to use the EC terms). I will close this PR and send some others regarding this  module in following days. Good luck!

PR #67, opened 2018-01-23
[FIX] runbot build on 11.0

gurneyalex:
the runbot build is 'orange' because of a warning issued
when an undocumented semi-dependency is not present. Check
that installing it cures the build


gurneyalex:
@pedrobaeza if this works then we can consider fixing MQT so that travis install nightly installs phonenumbers

pedrobaeza:
Please see instead this: https://github.com/OCA/OCB/pull/696

gurneyalex:
@pedrobaeza I'm fine with your suggested way too

I'd proposed https://github.com/OCA/maintainer-quality-tools/pull/527

PR #52, opened 2017-06-26
[MIG][10.0] account_cutoff_accrual_picking

TDu:
Hello,
Started migrating account_cutoff_accrual_picking, some schema changes impact this module.
It's a work in progress.
Still looking for a way to calculate them for a previous date correctly.
Cheers.

PR #45, opened 2017-01-24
[MIG][10.0] account_invoice_accrual migration

Cedric-Pigeon:
Depends on:

- [x] https://github.com/OCA/account-financial-tools/pull/445
- [x] https://github.com/OCA/account-closing/pull/43

PR #73, opened 2018-06-21
[MIG][10.0] account_cutoff_accrual_picking

jbaudoux:
This PR replaces https://github.com/OCA/account-closing/pull/52
I rebased to include changes of Alexis de Lattre in base modules and I finalized the work.
The module supports accrual based on SO and PO for goods.

The accrual is computed by comparing on the SO / PO, the quantity delivered/received and the quantity invoiced. In case, some moves or invoices have occurred after the cutoff date, those quantities can be affected and are recomputed. This allows to quickly generate a cutoff snapshot by processing few lines.

For PO, you can make the difference between:
- invoice to receive (received qty > invoiced qty)
- goods to receive (prepayment)  (received qty < invoiced qty)
If you expect a credit note, you can make it in draft. In standard, this update the PO and the quantity will not be accrued as goods to receive.
Thanks to the PR https://github.com/OCA/account-closing/pull/45
you can accrue the draft credit notes as "credit notes to receive".

For SO, you can make the difference between:
- invoice to generate (delivered qty > invoiced qty)
- goods to send (prepayment) (delivered qty < invoiced qty)
The credit notes are not affecting the invoiced quantity.


account_cutoff_accrual_picking/i18n/de.po
```diff
@@ -73,11 +196,36 @@ msgstr "Entnahme"
...
 #. module: account_cutoff_accrual_picking
-#: code:addons/account_cutoff_accrual_picking/account_cutoff.py:45
-#: code:addons/account_cutoff_accrual_picking/account_cutoff.py:66
-#: code:addons/account_cutoff_accrual_picking/account_cutoff.py:99
+#: code:addons/account_cutoff_accrual_picking/models/account_invoice.py:77
 #, python-format
+<<<<<<< 592ea12193282be4a0d2c7bbf637e625683bd609:account_cutoff_accrual_picking/i18n/de.po
 msgid "Error:"
 msgstr "Fehler:"
+=======
```
ghost: not resolver merge conflict!!!

account_cutoff_accrual_picking/i18n/de.po
```diff
@@ -73,11 +196,36 @@ msgstr "Entnahme"
 #: field:account.cutoff.line,picking_date_done:0
 msgid "Date Done of the Picking"
 msgstr "Datum erledigter Kommissionierung"
+=======
```
ghost: not resolver merge conflict!!!

account_cutoff_accrual_picking/i18n/de.po
```diff
@@ -60,11 +102,92 @@ msgstr "Fehlende '%s' bei Steuer '%s'."
 #, python-format
 msgid ""
 "Missing expense account on product '%s' or on its related product category."
+=======
```
ghost: not resolver merge conflict!!!

account_cutoff_accrual_picking/i18n/de.po
```diff
@@ -46,12 +62,38 @@ msgstr "Positionen aus Kommissionierung erneuern"
 #, python-format
 msgid ""
 "Missing income account on product '%s' or on its related product category."
+=======
```
ghost: not resolver merge conflict!!!

account_cutoff_accrual_picking/i18n/de.po
```diff
@@ -1,27 +1,43 @@
-# Translation of OpenERP Server.
+# Translation of Odoo Server.
 # This file contains the translation of the following modules:
+<<<<<<< 592ea12193282be4a0d2c7bbf637e625683bd609:account_cutoff_accrual_picking/i18n/de.po
 # 	* account_cutoff_accrual_picking
+=======
```
ghost: not resolver merge conflict!!!

account_cutoff_accrual_picking/i18n/de.po
```diff
@@ -1,27 +1,43 @@
...
 msgid ""
 msgstr ""
-"Project-Id-Version: OpenERP Server 7.0\n"
+"Project-Id-Version: Odoo Server 10.0+e\n"
 "Report-Msgid-Bugs-To: \n"
+<<<<<<< 592ea12193282be4a0d2c7bbf637e625683bd609:account_cutoff_accrual_picking/i18n/de.po
 "POT-Creation-Date: 2013-12-24 15:27+0000\n"
 "PO-Revision-Date: 2015-02-05 23:47+0100\n"
 "Last-Translator: Rudolf Schnapka <rs@techno-flex.de>\n"
+=======
```
ghost: not resolver merge conflict!!!

ghost (changes requested):
Lots of translation files contain not resolver merge conflict!!!

jbaudoux:
@oleksandrpaziuk I removed that wrong de.po

ghost:
@jbaudoux Is it okay to delete `de.po` file?

jbaudoux:
@oleksandrpaziuk It was quite obsolete

leemannd:
@jbaudoux I've not seen it. Normally we don't have to delete them.
@pedrobaeza Can we have your opinion on it?

pedrobaeza:
This also has Travis and runbot red

leemannd:
@jbaudoux You still have flake8 errors. Can you fix it?

jbaudoux:
@leemannd This PR is now green :-)

alexis-via:
For me, there is a major design issue in this PR. This design problem can be seen with the following scenario:
- create a PO dated 2018-09-03 for 10 keyboards
- receive 3 keyboards (stock move done on 2018-09-03)
- create a supplier invoice from the PO for the 3 keyboard and force the date of this invoice to 2018-09-05
- create an expense cutoff dated 2018-09-04: with the current design, it will not create a cutoff line for the 3 keyboards, although it should.

When you think about this and look at the code, you understand that it's a design problem. Porting account_cutoff_accrual_picking from v8 to v10 is not easy because of the major change in the datamodel and has to be thought carefully, taking into account all the scenarios.

I'm working on an implementation for v10.

jbaudoux:
Hello Alexis,

Thanks for your interest in this PR.
Your scenario does not make sense. How can you receive an invoice from a
supplier that is dated in the future?
A supplier invoice must be dated to today or in the past. And I ensured
that when you encode a supplier invoice on the 5 but dated on the 3, it
adjust the cutoff. This is a quite likely scenario.

Kind regards,
J-E


Le lun. 3 sept. 2018 à 20:33, Alexis de Lattre <notifications@github.com> a
écrit :

alexis-via:
Of course, it doesn't make sense to have an invoice in the future !!! But, when you TEST a scenario on Odoo, you cannot date your stock moves in the past. That's why I use this trick to show the problem.

The real-life scenario is the following:
- you have a PO dated 2017-12-15 for 10 keyboards
- you receive 3 keyboards on 2017-12-30
- you receive an invoice for 3 keyboards dated 2018-01-02
- on 2018-01-12, you generate your cutoff with cutoff date = 2017-12-31.
Expected result: Odoo should create a cutoff line for the 3 keyboards. Current result: it doesn't generate a cutoff line for the 3 keyboards.

jbaudoux:
Yes it does if you run your cutoff on 2017-12-31 midnight. That's why there
is a cron job doing this. Then when you encode the invoice in January, it
adapts the cutoff if the invoice is dated in December.
Nevertheless, there is idd a limitation that you cannot run a cutoff in the
past. If you find a solution that does not slow down stock move, nor
invoicing, I'll be happy to review.

Le lun. 3 sept. 2018 à 22:09, Alexis de Lattre <notifications@github.com> a
écrit :

alexis-via:
OK, I have a better understand of your implementation now. But I don't understand why you consider that we cannot generate cutoffs in the past without saving the state of order lines on the cutoff date. Order lines have O2M to invoice lines (with date) and move lines (with date), so we can know the qty of move before/after cutoff + invoice qty before/after cutoff.

I'll explain my idea with more details tomorrow when my proto starts working.

jbaudoux:
That's correct. You could theoretically solve it like that. But how do you
know at which moves and invoices to look in your computation? Where do you
stop in the past?
I found it easier to take a picture of sale order on cutoff date.

Le lun. 3 sept. 2018 à 22:40, Alexis de Lattre <notifications@github.com> a
écrit :

jbaudoux:
An additional advantage of doing like this, is that we do the computation
only based on purchase order which consolidate reception and invoicing. So
it could also maybe work for non stockable products (I don't know if there
is any use case?), but more important, it works even if you haven't
migrated stock moves

alexis-via:
For that, my starting point would be incoming/outgoing pickings with state == 'done' and with date_done <= cutoff_date... until date_done >= cutoff_date - 3 months (configurable). Then, the stock moves of those pickings are used to identify order lines to analyse, and then all the analysis is made from order lines (looking at related moves and related invoice lines).
The drawback of my design is that you must put a limit of the analysis in the past (cutoff_date - 3 months for example). But this drawback seems lighter to me than the need for a cron + inherit of invoice validation + unlink, which impacts day-to-day use of Odoo. For me, the major issue with your design is that you cannot generate cutoff if the cron has not been run on the cutoff date: if this didn't happen, then you're screwed!

alexis-via:
Here is my implementation of account_cutoff_accrual_picking for v10 https://github.com/OCA/account-closing/pull/84

jbaudoux:
rebased

jbaudoux:
I fixed the cutoff for revenues. You can now also generate the cutoff after the date (it will be slower as more computation is required).
A glue module account_cutoff_accrual_picking_product_analytic to be compliant with product_analytic has been added.
Many tests have been added.

PR #77, opened 2018-06-27
[10.0] account_multicurrency_revaluation: Fix tests using company from demo data

grindtildeath:

PR #78, opened 2018-06-27
[10.0][FIX] fix account_multicurrency_revaluation test

ghost:


guewen:
@oleksandrpaziuk can you base your work on #77 using a revaluation company? It seems the test pass only because of the change of the rate to 1.0, which is probably only hiding the issue.

fclementic2c:
@sbidoul better ?

ghost:
As I understand correctly test was made to test currency difference. And in the original test, we has EUR to USD comparison. After new changes in odoo, we can't change currency in the company so I swap currency and now we test USD to EUR with same rates. That's why test pass. If it is the wrong idea than I can use a new company with EUR to USD comparison like in #77

grindtildeath:
Please close in favour of #77

PR #79, opened 2018-07-03
[11.0] account_multicurrency_revaluation: Fix tests using company fro…

ghost:
…m demo data


account_multicurrency_revaluation/demo/account_demo.xml
```diff
@@ -1,66 +1,105 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
+    <record id="res_partner_reval" model="res.partner">
+        <field name="name">Revaluation Company</field>
+        <field eval="1" name="customer"/>
+        <field eval="1" name="supplier"/>
+        <field name="is_company">1</field>
```
asaunier: Just for my information: I guess you could have used
```
<field eval="1" name="is_company"/>
```
here, as for the lines around, correct? If yes, what's the recommended way?

account_multicurrency_revaluation/demo/account_demo.xml
```diff
@@ -1,66 +1,105 @@
...
+        <field eval="1" name="active"/>
+        <field name="street">90 Streets Avenue</field>
+        <field model="res.country" name="country_id" search="[('code','ilike','ch')]"/>
```
asaunier: Would it be possible to use ``base.ch`` instead?
```
<field name="country_id" ref="base.ch"/>
```
?

account_multicurrency_revaluation/tests/test_currency_revaluation.py
```diff
@@ -49,22 +52,31 @@ def setUpClass(cls):
         }
 
         partner = ref('base.res_partner_3')
+        partner.company_id = cls.company.id
+        partner.property_account_payable_id = receivable_acc.id
+        partner.property_account_receivable_id = payable_acc.id
+
+        payment_term = ref(
+            'account.account_payment_term')
```
asaunier: Do you really need a line break here?

grindtildeath:
@oleksandrpaziuk Instead of adding a new commit, couldn't you cherry-pick this commit https://github.com/OCA/account-closing/pull/77/commits/f1a32abbe436aae5532992dfbc4b64de801745b0 and add required changes in another commit so I can also cherry-pick your changes for 10.0 branch afterwards ?
Issue is we don't know what's the same between https://github.com/OCA/account-closing/pull/79/commits/2c882648fc361c33e55ffd59a187628874465c4d and the aforementioned one for 10.0, thus reviews would have to be made two times.

ghost:
@grindtildeath tests for 10 an 11 are a little bit different in method location but do the same job. and cherry-picking will rise a conflict. Should I replace tests from 11 with tests from 10 anyway?

grindtildeath (approved):
@oleksandrpaziuk Ok I didn't check it before, then no worries. 
Would have been better if this refactoring was done on both branch, but I don't have anymore time to loose on this so it's good to me, thank you. :wink:

grindtildeath (changes requested):
Did approve slightly too fast :(

Can you please also port this commit ? https://github.com/OCA/account-closing/pull/77/commits/90d2cca6c143988989cdd9ade41ec15f7a940cc8

ghost:
already renamed in 11

grindtildeath (approved):
@oleksandrpaziuk My bad then :sweat_smile:
Please also increase version number, and we're all good.

account_multicurrency_revaluation/tests/test_currency_revaluation.py
ghost: @asaunier Fixed

account_multicurrency_revaluation/demo/account_demo.xml
ghost: IMHO this is two method how to do the same things.
in this case, it is better to follow one style.

account_multicurrency_revaluation/demo/account_demo.xml
ghost: Yep. I think it is a better solution.

fclementic2c:
:+1:

PR #82, opened 2018-08-21
[10.0] account_invoice_accrual : Fix typo self._context

jbaudoux:


account_invoice_accrual/models/account_invoice.py
```diff
@@ -71,7 +71,7 @@ def invoice_validate(self):
 
     @api.multi
     def action_cancel(self):
-        if 'is_merge' not in self.context:  # See account_invoice_merge module
+        if 'is_merge' not in self._context:  # See account_invoice_merge module
```
sbidoul: `self.env.context` ? (I think self._context is deprecated)

jbaudoux: of course !

sbidoul:
Fast tracking obvious fix.

PR #83, opened 2018-08-24
[10.0][FIX] account-multicurrency-revaluation:  add hours to revaluation date as it is required to find out the correct rate

Cedric-Pigeon:
The revaluation wizard requests a data in input but currency rates are searched based on a DateTime.

If you try to revaluate on July 31, you will get the rate of July 30.
This because of this query: https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_currency.py#L43

The name of the currency rate is the date with hours in string

For example '2018/07/31'  is not higher than '2018/07/31 10:00:00'

This PR fixes the problem by adding '23:59:59' behind the wizard date.

### DIFF
--- a/account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py
+++ b/account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py
@@ -65,10 +65,10 @@ class WizardCurrencyRevaluation(models.TransientModel):
 
         # Compute unrealized gain loss
         ctx_rate = context.copy()
-        ctx_rate['date'] = form.revaluation_date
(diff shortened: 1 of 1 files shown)


fclementic2c (approved):

grindtildeath:
Just passing by, but shouldn't the default value here (https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_currency.py#L208) prevent such a behaviour ?

Setting the hour 10:00:00 (as in your example) on the rate is a manual action, and IMO it's wrong to consider it as the "day rate" as a currency can potentially change quite a lot during between 00:00:00 and 10:00:00, not to say 23:59:59. :confused:

PR #84, opened 2018-09-04
[10.0][MIG] account_cutoff_accrual_picking to v10

alexis-via:
This is a full re-write of the module account_cutoff_accrual_picking to port it from v8 to v10.

Use account_analytic_id instead of account_analytic_code in views because the 'code' field of analytic account is not a required field


jbaudoux:
Can you also consider the case where invoice qty != 0 but no picking yet? The are some real use cases: 
- prepayment. You received invoice before goods.
- transit. Supplier electronically sends the invoice at the same times as goods but goods are still in transit.
- delayed reception. Invoice is received but goods reception has not yet been encoded.

jbaudoux:
@alexis-via We should combine this with https://github.com/OCA/account-closing/pull/73. That latest one is very efficient if you have huge amount of stock moves. However if you don't generate on the cutoff date, yours is the way to go.

account_cutoff_accrual_picking/models/account_cutoff.py
```diff
@@ -0,0 +1,281 @@
...
+        if cutoff_type not in pick_type_map:
+            return res
+
+        # Create account mapping dict
+        account_mapping = acmo._get_mapping_dict(
+            self.company_id.id, cutoff_type)
+
+        # TODO date_done is a Datetime field, so maybe we need more clever code
+        # for our friends which are far away from GMT
+        pickings = spo.search([
```
jbaudoux, line 256: Technically you could have a move done without having the picking completely done. I suggest to analyze move instead of picking.
Functionally, I'm really concerned by the fact that the moves related to expenses/revenues can be of both type incoming and outgoing. For instance, for SO, you have outgoing moves for deliveries but incoming moves for returns.

jbaudoux (commented):
You start from the moves to find out the order lines concerned. However, you should also consider all cases where the order quantity is billed but the stock move is not yet done. I don't see how you manage those cases. You probably need to perform the same and take all invoice lines linked to order line linked to stock move and add those order lines in the set to process.

Issue #94, opened 2018-12-27
Error in creating databases with demo data


labaggio:
The odoo account module, in Payment methods, already inserts a record with id account_payment_method_manual_in:
https://github.com/odoo/odoo/blob/fe381f95b704085eb16b23ed74da7924ae0872e7/addons/account/data/account_data.xml#L85

 You are trying to insert a record with the same id, conflicting with the former.

https://github.com/OCA/account-closing/blob/fb78e8a2b84a7fd38d9883b039613aa6dd9af0f8/account_multicurrency_revaluation/demo/account_demo.xml#L128

pedrobaeza:
The XML-ID is composed by the module name + the id, so no collapse. Any specific problem apart from observing the same id?

labaggio:
When I try to install this module into  a database  with  "Load demonstration data" checked, I have this error:
![schermata del 2018-12-29 16-02-14](https://user-images.githubusercontent.com/25666721/50539585-26bc3280-0b83-11e9-81f7-9f7234a0928f.png)
I'm doing other tests because it does not happen if I install the single module..

pedrobaeza:
Fixed in #95

PR #95, opened 2018-12-31
[11.0][FIX] Error in creating databases with demo data

bodedra:
Issue reference: https://github.com/OCA/account-closing/issues/94

Before PR:

- Create xml record with same ID as account module and use it on test case.

After PR:

- Directly use xml ID from account module


pedrobaeza (approved):
Fast-tracking this

PR #96, opened 2019-01-07
[11.0][IMP] Move options to accounting settings

TDu:
Up to now the configuration options were displayed on the company view.
Here they are moved to the accounting settings.


grindtildeath (commented):
@TDu Is there any reason not to define the fields in `res.config.settings` as related to the fields on `res.company` ? :thinking: 
As I understand you wouldn't have to override `get_values` and `set_values` and it would do the job just as well.

fclementic2c:
I am in favor of moving these settings to the accounting settings instead of the company form.

TDu:
@grindtildeath I think that is the proper way to extend res.config.settings. But you are correct that using a related field may work as well for less code.

TDu:
@grindtildeath  Indeed the related field works, which was not the case pre Odoo 11 with transient model.  So it greatly simplifies the code :smile:

grindtildeath (approved):
:+1:

account_cutoff_accrual_base/views/res_config_settings.xml
```diff
@@ -0,0 +1,40 @@
...
+
+          <div class="col-xs-12 col-md-12 o_setting_box">
+            <div class="o_setting_left_pane">
+            </div>
+            <div class="o_setting_right_pane">
+              <label string="Accruals"/>
+              <div class="content-group">
+                <div class="row mt16">
+
+                  <label string="Account for Accrued Expense" class="col-md-3 o_light_label"/>
```
jcoux: I think you can use
`<label for="default_accrued_expense_account_id" class="col-md-3 o_light_label"/>`
to avoid to redifine the string and use the string defined into the model.
Same remark for 4 fields.

account_cutoff_base/views/res_config_settings.xml
```diff
@@ -0,0 +1,34 @@
...
+
+          <h2>Account Closing Cut-offs</h2>
+          <div name="account_closing_cutoffs" class="row mt16 o_settings_container" title="">
+            <div class="col-xs-12 col-md-12 o_setting_box">
+              <div class="o_setting_left_pane">
+              </div>
+              <div class="o_setting_right_pane">
+                <div class="content-group">
+                  <div class="row mt16">
+                    <label string="Default Cut-off Journal" class="col-md-3 o_light_label"/>
```
jcoux: Same remark here.

account_cutoff_prepaid/views/res_config_settings.xml
```diff
@@ -0,0 +1,33 @@
...
+      <field name="arch" type="xml">
+        <xpath expr="//div[@name='account_closing_cutoffs']" position="inside">
+          <div class="col-xs-12 col-md-12 o_setting_box">
+            <div class="o_setting_left_pane">
+            </div>
+            <div class="o_setting_right_pane">
+              <label string="Cutoff prepaid"/>
+              <div class="content-group">
+                <div class="row mt16">
+                  <label string="Account for Prepaid Expense" class="col-md-3 o_light_label"/>
```
jcoux: Same remark here.

jcoux (changes requested):
Little modifications asked.

account_cutoff_accrual_base/views/res_config_settings.xml
TDu: You are correct, so I changed it.
I had to make the label size a little bigger so it would not overflow as the strings defined on the model all start with 'Default' which seems kind of superfluous to me.

grindtildeath (commented):
Can you please remove the `api` import to make travis happy ?

TDu:
I remove the unused api import and squash the PR

grindtildeath (approved):

grindtildeath:
@jcoux Can you please update your review ?

PR #99, opened 2019-01-14
[MIG] account_cutoff_accrual_base: Migration to 12.0

rven:


jeroenev:
build seems to fail because account_cutoff_base wasn't found
it's merged now, restarted build
EDIT: Travis seems to pass
not sure why RunBot fails (don't see any usefull info on the page)
code seems fine

PR #100, opened 2019-01-24
[REF] account_multicurrency_revaluation: Set the revaluation date on …

keylor2906:
…the context when calling the revaluate_currency method on the wizard view, to correctly use the rate of the input date, because without this change, the rate used for the creation of the move and lines would be of the day of creation and not of the required date.

### DIFF
--- a/account_multicurrency_revaluation/tests/test_currency_revaluation.py
+++ b/account_multicurrency_revaluation/tests/test_currency_revaluation.py
@@ -161,7 +161,7 @@ class TestCurrencyRevaluation(SavepointCase):
             'revaluation_date': '%s-03-15' % fields.Date.from_string(
                 fields.Date.today()).strftime('%Y'),
             'journal_id': self.reval_journal.id,
-            'label': '[%(account)s] wiz_test',
+            'label': '[%(account)s %(rate)s] wiz_test',
         }
         wiz = wizard.create(data)
         result = wiz.revaluate_currency()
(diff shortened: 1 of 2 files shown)


moylop260 (approved):
:+1:
Just sign the CLA, please

moylop260:
@grindtildeath @mpanarin
What do you think about?

fclementic2c:
Wow... That is a big problem. 
@grindtildeath can you have a look please ?

grindtildeath (changes requested):
@keylor2906 I might be wrong but the revaluation date is used to set the context on the currency computation here : https://github.com/OCA/account-closing/blob/11.0/account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py#L140

Could this context key be missing somewhere ? In such case, can you please point where this context should be applied ?

In such case, would be great to add a test to highlight this change.

keylor2906:
@grindtildeath 

The context would be needed here https://github.com/OCA/account-closing/blob/11.0/account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py#L311, so the rate returned here https://github.com/OCA/account-closing/blob/11.0/account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py#L150 would be of the date on the wizard and not date of execute.
Also needed for the rate of the account move line label https://github.com/OCA/account-closing/blob/11.0/account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py#L327.

grindtildeath:
@keylor2906 Thanks for the explanation. 

Actually the returned rate is only used to format the label, so the computation was right but the label did show the wrong rate.

I opened #101 to fix this with a test, hope it's ok for you. :relaxed:

keylor2906:
@grindtildeath Yes, I think it would be better to use https://github.com/OCA/account-closing/pull/101 instead of this fix.

PR #101, opened 2019-01-30
[11.0] account_multicurrency_revaluation: Fixes and improvements

grindtildeath:
Thanks to @keylor2906 for pointing this out.

Supersedes : #100


grindtildeath:
@hbto @moylop260 @bodedra @yvaucher @mpanarin  @matthieuchoplin 
Can you guys please pass by for a review to get this merged ?

https://github.com/OCA/account-closing/pull/102 and https://github.com/OCA/account-closing/pull/103 would enjoy your approval as well :wink:

bodedra (approved):
LGTM (code review)

i-vyshnevska (approved):

account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py
```diff
@@ -129,15 +129,15 @@ def _compute_unrealized_currency_gl(self, currency_id, balances, form):
...
 
-        currency = currency_obj.browse(currency_id)
+        currency = currency_obj.browse(currency_id).with_context(ctx_rate)
 
         foreign_balance = adjusted_balance = balances.get(
             'foreign_balance', 0.0)
         balance = balances.get('balance', 0.0)
         unrealized_gain_loss = 0.0
         if foreign_balance:
             ctx_rate['revaluation'] = True
```
TDu: By moving the assignation of the context up this line would have no effect, no ?

TDu (commented):
Also I am not sure what is the difference between applying the context on the browse instead of on the compute ?

account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py
grindtildeath: In such case the context is defined on `currency` var, so that it's not only used on the `compute` call but also when we get the rate here : https://github.com/OCA/account-closing/pull/101/files/3aaae75184fb6c3487270c4f23de8a1008eda73d#diff-2f2adec7633819d0a28eb35fef7fca93R150

TDu: Yes, but does changing the variable used for context after it has been assigned actually impacts the context ?
I don't think so.

grindtildeath: Sorry for the confusion but you are right.
However, as I looked into this context's usage, it was actually not needed, and I added the following commit to remove it.

moylop260 (approved):

account_multicurrency_revaluation/wizard/wizard_currency_revaluation.py
TDu: Simplified code then :smile:

TDu (approved):

PR #104, opened 2019-02-18
[10][IMP]account_multicurrency_revaluation report currency unrealized

Tonow-c2c:
- Change the grouping rule for bank account in the report
  -  The result for not in ['payable', 'receivable'] should be printed only the total 
     - with _Curr. Balance YTD_ and _Revaluation Rate_ only if they are consistent

- Changing report name from _Unrealized currency report_ to _Currency revaluation report_

#### Before
image

#### After

image

-----

### Menuitem changing 

- Moving to the right _Currency Unrealized_ menu item (on the same indent as _Aged Partner Balance_) (so in the _OCA accounting reports_ category )

- Changing _Currency Unrealized_ menu item name to _Currency revaluation_

#### Before
image

#### After
image

### DIFF
--- a/account_multicurrency_revaluation/report/currency_unrealized_report.py
+++ b/account_multicurrency_revaluation/report/currency_unrealized_report.py
@@ -4,6 +4,8 @@
 
 from odoo import api, models
 
+FLOAT_DIGIT = 2
+
 
 class ShellAccount(object):
 
@@ -22,9 +24,18 @@ class ShellAccount(object):
         self.keys_to_sum = ['gl_foreign_balance', 'gl_currency_rate',
                             'gl_revaluated_balance', 'gl_balance',
                             'gl_ytd_balance']
+        # to check if 'currency' and the 'revaluation rate'
+        # is the same for all lines
+        self.keys_to_check = ['curr_name', 'gl_currency_rate']
 
     def __contains__(self, key):
         return hasattr(self, key)
+
+    def _format_float(self, val, ndigits=FLOAT_DIGIT):
+        val_formated = val
+        if isinstance(val, float):
+            val_formated = "%.2f" % round(val, ndigits=ndigits)
+        return val_formated
 
     def get_lines(self):
         """Get all line account move line that are need on report for current
@@ -59,11 +70,32 @@ class ShellAccount(object):
     def compute_totals(self):
         """Compute the sum of values in self.ordered_lines"""
         totals = dict.fromkeys(self.keys_to_sum, 0.0)
+        checks = dict.fromkeys(self.keys_to_check)
         for line in self.ordered_lines:
             for tot in self.keys_to_sum:
                 totals[tot] += line.get(tot, 0.0)
+            for check in self.keys_to_check:
+                if checks.get(check) is None:
+                    checks[check] = set([line.get(check)])
+                else:
+                    checks[check].add(line.get(check))
         for key, val in totals.iteritems():
-            setattr(self, key + '_total', val)
+            val_formated = self._format_float(val)
+            setattr(self, key + '_total', val_formated)
+        for key, val in checks.iteritems():
+            if len(val) == 1:
+                check_value = val.pop()
+                check_value_formated = self._format_float(check_value)
+                setattr(self, key + '_total_check', check_value_formated)
+            else:
+                setattr(self, key + '_total_check', False)
+
+    def format_ordered_lines(self):
+        """ To render only float with 2 digits
+        """
+        for line in self.ordered_lines:
+            for key, val in line.iteritems():
+                line[key] = self._format_float(val)
 
 
 class CurrencyUnrealizedReport(models.AbstractModel):
@@ -83,6 +115,8 @@ class CurrencyUnrealizedReport(models.AbstractModel):
                 docs |= account
                 shell_accounts[account.id] = acc
                 acc.compute_totals()
+                acc.format_ordered_lines()
+
         docargs = {
             'doc_ids': docs.ids,
             'doc_model': 'account.account',
--- a/account_multicurrency_revaluation/report/currency_unrealized_report.py
+++ b/account_multicurrency_revaluation/report/currency_unrealized_report.py
@@ -2,7 +2,7 @@
 # Copyright 2012-2017 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo import api, models
+from odoo import api, models, tools
 from odoo.addons import decimal_precision as dp
 
 
@@ -23,9 +23,6 @@ class ShellAccount(object):
         self.keys_to_sum = ['gl_foreign_balance', 'gl_currency_rate',
                             'gl_revaluated_balance', 'gl_balance',
                             'gl_ytd_balance']
-        # to check if 'currency' and the 'revaluation rate'
-        # is the same for all lines
-        self.keys_to_check = ['curr_name', 'gl_currency_rate']
 
     def __contains__(self, key):
         return hasattr(self, key)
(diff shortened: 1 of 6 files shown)


account_multicurrency_revaluation/report/currency_unrealized_report.py
```diff
@@ -4,6 +4,8 @@
 
 from odoo import api, models
 
+FLOAT_DIGIT = 2
```
i-vyshnevska: use instead
self.env['decimal.precision'].precision_get('Your currency precision')

account_multicurrency_revaluation/report/unrealized_currency_gain_loss.xml
```diff
@@ -39,38 +39,50 @@
...
+                                                    <span t-esc="sh_acc.gl_foreign_balance_total or 0.0"/>
+                                                </b>
+                                            </t>
+                                        </div>
                                         <div class="act_as_cell"></div>
-                                        <div class="act_as_cell"></div>
-                                        <div class="act_as_cell"></div>
+                                        <div class="act_as_cell amount">
+                                            <b>
+                                                <span t-esc="sh_acc.gl_currency_rate_total_check or 'Revaluation Rate not homogeneous'"/>
```
i-vyshnevska: no need to show this in total, it has value only on journal items in total it would be misled if currency has changed in period

account_multicurrency_revaluation/report/currency_unrealized_report.py
```diff
@@ -22,10 +24,19 @@ def __init__(self, account):
         self.keys_to_sum = ['gl_foreign_balance', 'gl_currency_rate',
                             'gl_revaluated_balance', 'gl_balance',
                             'gl_ytd_balance']
+        # to check if 'currency' and the 'revaluation rate'
+        # is the same for all lines
+        self.keys_to_check = ['curr_name', 'gl_currency_rate']
```
i-vyshnevska: no need for such check, report is formed on journal and it's already on one currency

account_multicurrency_revaluation/report/unrealized_currency_gain_loss.xml
```diff
@@ -39,38 +39,50 @@
...
-                                                <span t-esc="line.get('gl_ytd_balance') or 0.0"/>
-                                            </div>
-                                        </div>
+                                        </t>
                                     </t>
                                     <div class="act_as_row lines labels">
                                         <div class="act_as_cell"><b>TOTAL</b></div>
+                                        <div class="act_as_cell amount">
+                                            <t t-if="account.internal_type not in ['payable', 'receivable'] and sh_acc.curr_name_total_check">
+                                                <b>
```
simahawk: use `<strong />` instead of b+span

simahawk (changes requested):
```bash
ParseError: "External ID not found in the system: account_financial_report_qweb.menu_oca_reports" while parsing /home/travis/build/OCA/account-closing/account_multicurrency_revaluation/wizard/print_currency_unrealized_report_view.xml:36, near
```

account_multicurrency_revaluation/report/currency_unrealized_report.py
Tonow-c2c: Hello I thinks I should check id the currency is consistent cause if the account have no `currency_id` so the constraint is applied  [https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_move.py#L445](https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_move.py#L445)

@fclementic2c  am I wrong ?

i-vyshnevska: and constraint on which you point should breakup entry on creation for different currencies

Tonow-c2c: even if I have a account without Account currency no ?
image

i-vyshnevska: yes " if not line.currency_id " - it's  when during work they start using currency, but moves already exist without currency, 
at least I had a lot of pain when we tried to mixin it

account_multicurrency_revaluation/report/currency_unrealized_report.py
```diff
@@ -22,10 +23,20 @@ def __init__(self, account):
...
 
     def __contains__(self, key):
         return hasattr(self, key)
 
+    def _format_float(self, val):
+        ndigits = dp.get_precision('Account')(self.cursor)[1]
+        val_formated = val
+        if isinstance(val, float):
+            val_formated = "%.2f" % round(val, ndigits=ndigits)
```
simahawk: any reason not to use `tools.float_round` ?

account_multicurrency_revaluation/report/currency_unrealized_report.py
```diff
@@ -59,11 +70,32 @@ def get_lines(self):
     def compute_totals(self):
         """Compute the sum of values in self.ordered_lines"""
         totals = dict.fromkeys(self.keys_to_sum, 0.0)
+        checks = dict.fromkeys(self.keys_to_check)
```
simahawk: pls do not use `dict` constructor use `{}.fromkeys`

Tonow-c2c: that was to coherent with the line above

account_multicurrency_revaluation/report/currency_unrealized_report.py
sebalix: I'm not sure that the constraint is applied if we have no currency on the account here (`if line.account_id.currency_id:`), in this case no `ValidationError` is raised and we can mix currencies.

i-vyshnevska: I've check it, if we have no currency for account and trying create lines with currency we get error, so while you group lines under one account no need to worry about multiple currency

i-vyshnevska (approved):
LG,  just check report when many pages formed

account_multicurrency_revaluation/report/currency_unrealized_report.py
```diff
@@ -26,6 +27,13 @@ def __init__(self, account):
     def __contains__(self, key):
         return hasattr(self, key)
 
+    def _format_float(self, val):
+        ndigits = dp.get_precision('Account')(self.cursor)[1]
+        val_formated = val
+        if isinstance(val, float):
+            val_formated = "%.2f" % tools.float_round(val, ndigits)
```
mpanarin: You are doing the double rounding here. You round in your format with `%.2f` as well as with `flat_round`. Keep the latter.

mpanarin (changes requested):

account_multicurrency_revaluation/report/currency_unrealized_report.py
Tonow-c2c: @mpanarin  the first is for the value the second is for the printing 
even if I don't do :
 `"%.2f"` 1.00 is printed like 1.0

Tonow-c2c: I am in this case @mpanarin 
```python
  >>> 125650429603636838/(2**53)
  13.949999999999999

  >>> 234042163/(2**24)
  13.949999988079071

  >>> a=13.946
  >>> print(a)
  13.946
  >>> print("%.2f" % a)
  13.95
  >>> round(a,2)
  13.949999999999999
  >>> print("%.2f" % round(a,2))
  13.95
```
https://stackoverflow.com/questions/455612/limiting-floats-to-two-decimal-points

simahawk: https://github.com/odoo/odoo/blob/10.0/odoo/tools/float_utils.py#L136
```py
def float_repr(value, precision_digits):
    """Returns a string representation of a float with the
       the given number of fractional digits. This should not be
       used to perform a rounding operation (this is done via
       :meth:`~.float_round`), but only to produce a suitable
       string representation for a float.
        :param int precision_digits: number of fractional digits to
                                     include in the output
    """
```

Tonow-c2c: thanks @simahawk

mpanarin (approved):

sebalix (approved):

simahawk (approved):

Issue #107, opened 2019-03-29
cutoff rounding


sbidoul:
We have had a case of unbalanced bs/pl accounts due to the cutoff module. 

I think it should [apply rounding here](https://github.com/OCA/account-closing/blob/6c0a8956bf6b4b565c8c5c92671661f5093370e5/account_cutoff_accrual_dates/models/account_cutoff.py#L44-L45).

Otherwise we end up with accounting entries with many decimal which cause issues down the line.

@alexis-via what do you think?

sbidoul:
cc/ @chrisdec

gaikaz:
It looks like we encountered this issue as well on Odoo 12.0.
@sbidoul Any more intel on this after a year?
Seems like your mentioned code is still there in v13

alexis-via:
@gaikaz Do you have the proof that the bug still exists in v13 ? The rounding is not done in the cutoff line, but when generating the account.move.line, and this code is in account_cutoff_base, and, for me, the fix is included in v13 (cf calls for round() in account_cutoff_base in v13).

gaikaz:
No, we only just found it on v12.
Today a coworker will try to reproduce this on the Community Runbot with as little configuration as possible. If it works, we'll check it on v13.

In short, I will write here about anything she finds.

alexis-via:
Did you find it on v12 with up-to-date code ? It would be very surprising, because the fix is also present in v12.

gaikaz:
Indeed, our client was still running older code.
We kind of wasted a little time now. I compared the versions and they were the same. Why was the version not bumped on PR merges? I trusted that we have the newest one on the client server :stuck_out_tongue: 

I think this issue can be closed, though.

gaikaz:
@alexis-via 
I thank you for the quick responses 😊
We didn't have any more trouble anymore.

PR #109, opened 2019-04-05
[10.0] FIX bug rounding debit/credit

alexis-via:
In the DB, the debit or credit should look like 244.25 or 244.45000000000005, but should NEVER look like 244.45770338474394 because it leads to annoying problems in accounting reports. For that, we need to round before doing the create of account.move.line in DB.

This is a very important bug, but leads to total debit != total credit in a balance ! More explanations to come.
Maybe we can improve this fix to also round in cutoff lines.

### DIFF
--- a/account_cutoff_accrual_dates/models/account_cutoff.py
+++ b/account_cutoff_accrual_dates/models/account_cutoff.py
@@ -5,7 +5,6 @@
 
 from odoo import models, fields, api, _
 from odoo.exceptions import UserError
-from odoo.tools import float_round
 
 
 class AccountCutoff(models.Model):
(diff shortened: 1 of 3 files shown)


alexis-via:
To fix your account move lines impacted by this bug, you can use this code: https://github.com/akretion/odoo-usability/blob/10.0/account_usability/account.py#L377

sbidoul:
See also #107

alexis-via:
@sbidoul I confirm this BIG FAT PROBLEM ; it's really a pain. It seems we discovered the same problem at customers sites almost at the same time. Did you try my script that fixes data above ?
I think we should have an important communication in the OCA about this issue, because all developers that create account.move.lines should be aware of this potential problem. The same bug can be found in account_move_template ; I fixed it in my port to v12 (https://github.com/OCA/account-financial-tools/pull/801). I had a similar bug in my module account_vat_pro_rata (https://github.com/akretion/odoo-vat-prorata) ; I already fixed it there.

alexis-via:
This PR is now ready for review+merge.
We also need to communicate in OCA to make sure that:
1) developers take this problem into account when creating account.move.lines
2) people impacted by this issue with cutoff modules or account_move_template can fix their data (with my script https://github.com/akretion/odoo-usability/blob/10.0/account_usability/account.py#L377 or something else)

sbidoul (approved):

PR #122, opened 2019-10-30
[IMP] Allow to split by partner + group by tag group

vrenaville:
* Purpose
Sometimes we want to split move line by partner, I have add a boolean to enable this feature.
It also group by analytic tag group


grindtildeath (changes requested):
Change looks ok, however I guess we must have an option to enable/disable the grouping by `analytic_tag_ids` for retro compatibility, since it wasn't there before...

Moreover, can you please bump version number in `__manifest__.py` ?

PR #128, opened 2020-02-19
[RFR] [FIX] 12.0 remove duplicate demo data entry from account_multicurrency_revaluation

PieterPaulussen:
Unnecessary duplicate demo data entry which is already present in odoo's account module:
https://github.com/odoo/odoo/blob/12.0/addons/account/data/account_data.xml#L100-L104

Also refer to the appropriate xml id in the test.


StefanRijnhart (approved):
Got hit by the same problem. Violates https://github.com/OCA/bank-payment/blob/12.0/account_payment_mode/models/account_payment_method.py#L35-L39 in my CI (which is an entirely valid constraint).

PR #129, opened 2020-03-21
12.0 mig account fiscal year closing

sergiocorato:


MarcoCalcagni:
Hi Sergio , 

I test the PR and I find and error in account_fiscal_year_closing/models/account_fiscalyear_closing.py on line 519 it should be 
`        if config.move_id:
            move_vals = config.move_id.reverse_moves(
                date=date, journal_id=self.journal_id,
            )
`
insted of 
`        if config.move_id:
            move_vals = config.move_id.reverse_moves(
                date=date, journal=self.journal_id,
            )
`
this happens  when the Inverse config is set up for the opening account move . 
also 
`    def moves_create(self):
        self.ensure_one()
        moves = self.env['account.move']
        # Prepare one move per configuration
        data = False
        if self.mapping_ids:
            move_lines = self._mapping_move_lines_get()
            data = self.move_prepare(move_lines)
        elif self.inverse:
            data = self.inverse_move_prepare()
            return True, data
`
I add the last row because the following check I believe  are not necessary . 
the opening account move  have a wrong description and is also validated . I don't know how to fix. 
Last the button "confirm and post move" generate an error. 

Ciao Marco

sergiocorato:
Ciao @As400it 
thanks, I committed a fix for inverse

MarcoCalcagni:
> Ciao @As400it
> thanks, I committed a fix for inverse

Ciao Sergio,
2 things . 

1) if you check the reconcile i got an error 
image

in account_fiscalyear_closing.py on row 349 for  config.move_id.move_reverse_reconcile() 
AttributeError: 'account.move' object has no attribute 'move_reverse_reconcile'

2) the opening move have a wrong reference note .
image
i believe it should be the description of config row 

Marco

kittiu:
@sergiocorato thank you for bringing this up.
Do you plan to include test script of this module too?

sergiocorato:
@kittiu yes, I'll try

kittiu:
> @kittiu yes, I'll try

Thank you! Will be happy to review :) Please ping me

cicciodeg:
Hi guys, I can reproduce all Marco's bugs plus one (the n.4):

1. The reference name of Opening moves is "Storno di :/" regardless of how I set the Description (of the Moves configuration in Fiscal year closings)
2. Opening moves are immediately confirmed
3. If I set reconcile=true on the Opening moves, I cannot change the status of the Fiscal year closings into Posted because when I click on the button "Confirm and post moves" I get this error 

```
[...]
File "/opt/odoo-addons/account-closing/account_fiscal_year_closing/models/account_fiscalyear_closing.py", line 349, in button_post
    config.move_id.move_reverse_reconcile()
AttributeError: 'account.move' object has no attribute 'move_reverse_reconcile'

```

4. If I set reconcile=false on the Opening moves, I CAN change the status to Posted by clicking on "Confirm and post moves" but the generated moves have an inverted numeration... in the example, as you can see the first confirmed registration of 2018 should have been "Epilogo ricavi" and not "Chiusura patrimoniale - passivo"

Documento2

andreampiovesana:
rename PR with [12.0][MIG] ?
very good work!

sergiocorato:
@cicciodeg @kittiu @As400it I updated fixing move sequence post, name of reverse move, and reconcile by removing it (even if boolean is not removed - I'll do it later) as Odoo core function do the reconciliation by default.
Anyway I think this option to do not reconcile it's unuseful, but if someone has a valid opinion to restore it, I'll try to do.

MarcoCalcagni (approved):
Hi Sergio, its ok for me.

andreampiovesana (approved):
go

sergiocorato:
With the last commits I cleaned the field `reconcile` which has no effect, regenerated README, re-enabled configuration of move date on fyc template, and cleaned/fixed tests.

sergiocorato:
I started #134 to fix accounting report that will not work properly with closing moves. Every info of other reports with similar issues will be welcome :D

sergiocorato:
@OCA/accounting-maintainers something is going wrong with ocabot

pedrobaeza:
Sometimes it doesn't work, also because GitHub has been very unstable lately. Just repeat the command.

PR #134, opened 2020-04-25
[12.0][ADD] glue module for account fiscal year closing reporting

sergiocorato:
Using #129 will create moves that are shown in accounting report.
This module prevents it for General ledger report, and for other in WIP.

Depends on #129


account_fiscal_year_closing_reporting/models/general_ledger_report_compute.py
```diff
@@ -0,0 +1,375 @@
...
+               AND m.closing_type not in ('closing', 'opening', 'loss_profit')
+                """
+                % ml_date
+            )
+        return sub_subquery_sum_amounts
+
+    # this function is changed only adding lines:
+    # AND
+    # m.closing_type not in ('closing', 'opening', 'loss_profit')
+    def _inject_line_not_centralized_values(
```
sergiocorato, line 54: @JordiBForgeFlow @pedrobaeza @eLBati I didn't found a way to inherit this function to do [this](https://github.com/OCA/account-closing/pull/134/files#diff-882b756955caf29ae168b87fdc00d98aR308-R309) simple change to limit the account moves to be considered, so I do this ugly copy... If there is a better way, it'll be welcome!

eLBati:
@sergiocorato it is probably worth of  reviewing https://github.com/OCA/account-financial-reporting/pull/660 and basing this PR on that new refactor, where indeed `_inject_line_not_centralized_values` is removed.
What do you think?

sergiocorato:
> @sergiocorato it is probably worth of reviewing [OCA/account-financial-reporting#660](https://github.com/OCA/account-financial-reporting/pull/660) and basing this PR on that new refactor, where indeed `_inject_line_not_centralized_values` is removed.
> What do you think?

:100:

eLBati:
Forget it: https://github.com/OCA/account-financial-reporting/pull/660#issuecomment-621273250

account_fiscal_year_closing_reporting/models/general_ledger_report_compute.py
eLBati: So, at the moment, I can't see other ways to do this

eLBati:
@sergiocorato trying to follow a more simple approach and to avoid this big override, we could use a particular journal for closing entries and then automatically exclude it from `journal_ids` in trial balance wizard (possible after https://github.com/OCA/account-financial-reporting/pull/678).
What do you think?

sergiocorato:
> @sergiocorato trying to follow a more simple approach and to avoid this big override, we could use a particular journal for closing entries and then automatically exclude it from `journal_ids` in trial balance wizard (possible after [OCA/account-financial-reporting#678](https://github.com/OCA/account-financial-reporting/pull/678)).
> What do you think?

@eLBati as discussed directly, this approach would lead to a complicated checking of the databases in all existing installation from users (create and set a specific journal for closing, recreate closing on this journal, etc.), so would be very time-expensive.

eLBati:
OK. I haven't had time to test this so far. But I'm OK with the approach

sergiocorato:
> OK. I haven't had time to test this so far. But I'm OK with the approach

@eLBati approve?

eLBati:
@sergiocorato  fix  error at installation: https://github.com/sergiocorato/account-closing/pull/1

account_fiscal_year_closing_reporting/models/general_ledger_report_compute.py
eLBati: @sergiocorato could you just add a comment to the modified lines ?

sergiocorato:
> @sergiocorato fix error at installation: [sergiocorato#1](https://github.com/sergiocorato/account-closing/pull/1)

I don't see the error, but ok for me

sergiocorato:
@eLBati can this be merged?

eLBati:
Ciao @sergiocorato see https://discord.com/channels/753902328494424064/753902328494424070/814901061361729586 so I am abstaining

sergiocorato:
Closing as unused.

PR #136, opened 2020-06-18
[12.0] FIX account_cutoff_accrual_dates: compute Accrual On Taxes only if enabled

eLBati:
and

account_cutoff_accrual_base: default value not set due to wrong context value

### DIFF
--- a/account_cutoff_accrual_base/models/account_cutoff.py
+++ b/account_cutoff_accrual_base/models/account_cutoff.py
@@ -12,7 +12,7 @@ class AccountCutOff(models.Model):
     @api.model
     def _default_cutoff_account_id(self):
         account_id = super(AccountCutOff, self)._default_cutoff_account_id()
-        cutoff_type = self.env.context.get('default_cutoff_type')
+        cutoff_type = self.env.context.get('cutoff_type')
         company = self.env.user.company_id
         if cutoff_type == 'accrued_expense':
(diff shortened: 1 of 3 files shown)


account_cutoff_accrual_base/models/account_cutoff.py
```diff
@@ -12,7 +12,7 @@ class AccountCutOff(models.Model):
     @api.model
     def _default_cutoff_account_id(self):
         account_id = super(AccountCutOff, self)._default_cutoff_account_id()
-        cutoff_type = self.env.context.get('default_cutoff_type')
+        cutoff_type = self.env.context.get('cutoff_type')
```
sergiocorato, line 15: Sure is this the problem? I find other functions with this logic, Travis is red too

eLBati: `default_cutoff_type` is never set in context.

I will check travis.

Thanks

eLBati: Fixed

sergiocorato (approved):

account_cutoff_accrual_base/models/account_cutoff.py
sergiocorato: :+1:

marcelsavegnago (approved):

PR #146, opened 2020-10-20
11.0 mig account fiscal year closing

bodi000:
Added postgres to travis as checks were failing.
MIG of account_fiscalyear_closing to v11


account_fiscal_year_closing/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2016 Tecnativa - Antonio Espinosa
+# Copyright 2016-2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Fiscal year closing",
+    "summary": "Generic fiscal year closing wizard",
+    "version": "11.0.1.0.0",
+    "category": "Accounting & Finance",
+    "website": "https://www.tecnativa.org/",
```
MeetKD, line 10: website -> https://github.com/OCA/account-closing/

account_fiscal_year_closing/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+
```
MeetKD, line 1: Empty line can be removed and license terms ?

account_fiscal_year_closing/__manifest__.py
bodi000: Thank you. I would ask @pedrobaeza what does he think, not a specialist in these matters myself.

account_fiscal_year_closing/models/__init__.py
bodi000: Thank you. I would ask @pedrobaeza what does he think, not a specialist in these matters myself.

MeetKD: No Problem :+1:

account_fiscal_year_closing/__manifest__.py
pedrobaeza: Yes, it's OK to change to GitHub OCA websites :)

sergiocorato:
As in v. 12.0 this module is already ported, have you checked compatibility?

matteoopenf:
it there problem to merge that pr?

sergiocorato:
It would be better to check if the module in v. 12.0 has something different that might require a migration function, or check if this PR is equivalent to a backport from v. 12.0 so that the migration function is useless

matteoopenf:
> It would be better to check if the module in v. 12.0 has something different that might require a migration function, or check if this PR is equivalent to a backport from v. 12.0 so that the migration function is useless

@bodi000  Can you help us  to understand the problem?

PR #147, opened 2020-10-28
[10.0][FIX] account_multicurrency_revaluation: fixed bug with rate of curre…

jesus01x:
Proposed changes
This change fix the wrong label of revaluate adjust balance


alexey-pelykh:
@jesus01x any details how exactly wrong it is? What is the result right now and what is the expected result?

PR #149, opened 2020-12-23
[12.0][IMP] account_multicurrency_revaluation

alexey-pelykh:
Things done:
 * [FIX] #126, #117: bank account revaluations are not grouped by partner
 * [FIX] Bank account revaluations are not performed if currency of account matches company currency
 * [IMP] When revaluation is performed in the past, force-execute reversal (if needed)
 * [IMP] Exchange rate is rounded during representation in same way as Odoo does
 * [IMP] Float comparison operations take currency rounding into account
 * [FIX] README cleanup

PR #151, opened 2021-01-17
[12.0][FIX] account_fiscal_year_closing: tests should not use fixed dates

alexey-pelykh:


alexey-pelykh:
@hbto @pedrobaeza can we fast-track this so I can rebase #149 ?

sergiocorato (approved):
LGTM

pedrobaeza:
Why are you changing this? Tests must use fixed data, and not variable one. Any problem you think on it's not due to this.

alexey-pelykh:
@pedrobaeza well, w/o this change tests just fail 🤷‍♂️ With change for dates to be "this year", tests don't fail. I'm no expert in year closing, but that's as close as I could get, and last year (when fixed dates were "this year"), tests were not failing as well.

sergiocorato:
If I had time when I ported it, I would have done it myself. BTW, if @alexey-pelykh resolved an issue in another module with this change, perhaps there is something wrong somewhere.

alexey-pelykh:
@sergiocorato w/o this change, this specific module tests failed on original 12.0 branch, so nothing to do with other modules I guess

pedrobaeza:
The usual problem with variable dates comes with leap years not returning the same expected results, but in general, is not good to not control input data for having always the same output.

On the other hand, fixing dates can make that methods that uses date references like `today`, `now` and so on to not be testable. Solution is to mock the function to return a fixed date, and in v14, you can use `freezegun` library, as included in Odoo core.

sergiocorato:
@pedrobaeza Perhaps I understand your position, but as you see in the code, @alexey-pelykh is using fixed dates with `month` and `day`, not `today` as-is. Variability added is only regarding the year.

pedrobaeza:
OK, it was just a suggestion for keeping things with good guidelines, but there are tests with relative dates all around OCA, and they occasionally blown away, but others not, so we will know if it happens with this one.

PR #165, opened 2021-04-30
[12.0][IMP] closing type not null

sergiocorato:
By default `closing_type` is `'none'` but user can manually set null. Having a null value is bad for reporting, so this PR will remove this possibility.

### DIFF
--- a/account_fiscal_year_closing/__manifest__.py
+++ b/account_fiscal_year_closing/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Fiscal year closing",
     "summary": "Generic fiscal year closing wizard",
-    "version": "12.0.1.1.1",
+    "version": "12.0.1.2.0",
     "category": "Accounting & Finance",
     "website": "https://github.com/OCA/account-closing",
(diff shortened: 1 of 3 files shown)


account_fiscal_year_closing/views/account_move_views.xml
```diff
@@ -9,7 +9,7 @@
         <field name="inherit_id" ref="account.view_move_form"/>
         <field name="arch" type="xml">
             <field name="ref" position="after">
-                <field name="closing_type"/>
+                <field name="closing_type" required="True"/>
```
SimoRubi: This is redundant since the field is already defined as required, can you remove it?

SimoRubi (changes requested):
Thanks for the PR!
Just code review

account_fiscal_year_closing/views/account_move_views.xml
sergiocorato: Done, thanks!

SimoRubi (approved):
Just code review, ok for me

matteoopenf (approved):

PR #177, opened 2021-08-05
[14.0][FIX] account_cutoff_base: Move domains where they are used

rousseldenis:
Avoid ValidationError on res.company views update as company_id field
does not exist there


rousseldenis:
@sbidoul @alexis-via

sbidoul (approved):

alexis-via (approved):

PR #187, opened 2021-12-03
[13.0][IMP]account_invoice_start_end_dates: removed required on views

OriolVForgeFlow:
Removed the required on the view if the product had "must_have_dates" as it was deleting the account_move_lines (of products with mandatory start and end dates).
Now you can save the invoice but if you try to post it it's when a warning will appear to fill those mandatory fields (only if the product requires so) and won't post the invoice.
@JordiMForgeFlow @forgeflow


JoanMForgeFlow (approved):
Codee review!

mariadforgeflow (approved):
LGTM!

PR #213, opened 2022-12-05
[MIG] account_cutoff_base: Migration to 16.0

dzungtran89:
Needs
- https://github.com/OCA/account-closing/pull/212: for `pre-commit` fixes

Note:

- I removed the `groups` from the `company_id` field is to avoid violating the check by the native, see [here](https://github.com/odoo/odoo/blob/16.0/odoo/addons/base/models/ir_ui_view.py#L3014)


account_cutoff_base/models/account_cutoff_line.py
```diff
@@ -0,0 +1,79 @@
...
+    _name = "account.cutoff.line"
+    _inherit = "analytic.mixin"
+    _description = "Account Cut-off Line"
+
+    parent_id = fields.Many2one("account.cutoff", string="Cut-off", ondelete="cascade")
+    name = fields.Char("Description")
+    company_currency_id = fields.Many2one(
+        related="parent_id.company_currency_id",
+        string="Company Currency",
+        readonly=True,
```
ivantodorovich: readonly is redundant here, in case you want to remove it

account_cutoff_base/models/account_cutoff_line.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    partner_id = fields.Many2one("res.partner", string="Partner", readonly=True)
+    quantity = fields.Float(digits="Product Unit of Measure", readonly=True)
+    price_unit = fields.Float(
```
ivantodorovich, line 25: This could be a Monetary field

account_cutoff_base/models/account_cutoff_line.py
```diff
@@ -0,0 +1,79 @@
...
+        string="Unit Price w/o Tax",
+        digits="Product Price",
+        readonly=True,
+        help="Price per unit (discount included) without taxes in the default "
+        "unit of measure of the product in the currency of the 'Currency' field.",
+    )
+    price_origin = fields.Char(readonly=True)
```
ivantodorovich, line 32: I fail to understand the purpose of this field, and why it is a `Char` but it's name suggests a price (thus it should be `Float` or `Monetary`)

Is this field even used? If so, what for? Should this be commented in the `help` attribute?

account_cutoff_base/models/account_cutoff_line.py
```diff
@@ -0,0 +1,79 @@
...
+    origin_move_line_id = fields.Many2one(
+        "account.move.line", string="Origin Journal Item", readonly=True
+    )  # Old name: move_line_id
```
ivantodorovich: Maybe remove the `# Old name:` comments.
Seem like they're outdated by now

account_cutoff_base/models/account_cutoff_line.py
```diff
@@ -0,0 +1,79 @@
...
+    origin_move_id = fields.Many2one(
+        related="origin_move_line_id.move_id", string="Origin Journal Entry"
+    )  # old name: move_id
+    origin_move_date = fields.Date(
+        related="origin_move_line_id.move_id.date", string="Origin Journal Entry Date"
+    )  # old name: move_date
+    account_id = fields.Many2one(
```
ivantodorovich, line 42: IMO, this model should have `_check_company_auto = True`, and a check_company on this `account_id` field.
(Maybe also other fields, if relevant)

account_cutoff_base/models/account_cutoff_tax_line.py
```diff
@@ -0,0 +1,51 @@
...
+    _inherit = "analytic.mixin"
+    _description = "Account Cut-off Tax Line"
+
+    parent_id = fields.Many2one(
+        "account.cutoff.line",
+        string="Account Cut-off Line",
+        ondelete="cascade",
+        required=True,
+    )
+    tax_id = fields.Many2one("account.tax", string="Tax", required=True)
```
ivantodorovich: `check_company` maybe?

ivantodorovich (approved):
Non-blocking comments

dzungtran89:
@ivantodorovich , thank you for you comments, the PR has been updated

About the field `price_origin`, I'm not sure about its purpose tbh, it was initially added in the [commit](https://github.com/OCA/account-closing/pull/213/commits/a3fc49ed252697dc89bbf2cedc21535abaff89c5#diff-a30853050e1420bf74baaaba68936d3e32217851cc7aa701d50053ce3af9ee58R431)

account_cutoff_base/models/account_cutoff_line.py
alexis-via: No. price_unit are not monetary fields for decimal precision reasons, cf https://github.com/odoo/odoo/blob/16.0/addons/sale/models/sale_order_line.py#L139
A price_unit can have a decimal precision higher than the currency precision.

account_cutoff_base/models/account_cutoff_line.py
alexis-via: It's a small note field that indicate where the price has been taken from. So it contains an invoice number, a PO number or an SO number.

alexis-via:
I'm preparing a PR on this PR, to take into account my comments and make a few improvements

alexis-via:
Here it is: https://github.com/dzungtran89/account-closing/pull/1

alexis-via (approved):

simahawk:
@dzungtran89 next time, please, remove merge commits if you integrate other PRs. Thanks.

dzungtran89:
hi @simahawk, well noted with thanks

PR #223, opened 2023-01-12
[14.0] account_cutoff_base: store=True on related fields used in tree views

alexis-via:
account_cutoff_base: store=True on related fields used in tree views. Advantages:
- allows to sort cutoff lines on account code  (or group by on account code)
- faster loading of tree view of cutoff lines ?


nibraun (approved):
Successfully used in prod!

sergiocorato (approved):

PR #224, opened 2023-01-13
[14.0] subscription: re-implement the interval logic

alexis-via:
The interval now starts on the first fiscal year start date before cutoff date
If the interval is pro-rated, the minimum amounts and provisionning amounts are pro-rated too.
The goal of this important change is to improve the computation of cutoff for an intermediate financial statement (no real change when generating cutoffs for the full fiscal year)

I added a M2O to the subscription object on the cutoff line.


nibraun (approved):
Significative improvement for subscriptions, succesfully used in prod

PR #225, opened 2023-01-14
[14.0] account_cutoff_accrual_picking: add prepaid support

alexis-via:
The module should be renamed from account_cutoff_accrual_picking to account_cutoff_picking in v16, now that it has prepaid expense/revenue support.

### DIFF
--- a/account_cutoff_accrual_picking/__manifest__.py
+++ b/account_cutoff_accrual_picking/__manifest__.py
@@ -3,11 +3,11 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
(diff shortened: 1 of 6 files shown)


tv-openbig:
@alexis-via I have checked all four cut-off cases with proper accounts (i have used our typical german accounts for that). For all cases i get the expected results from an accountant perspective. Thanks a lot for adding the prepaid expenses / revenues.
Bildschirmfoto 2023-01-16 um 10 03 04

simahawk:
can we have some test coverage?

simahawk:
BTW you can rebase: I added GH actions to branch 14.0 here https://github.com/OCA/account-closing/pull/233

PR #226, opened 2023-01-16
[13.0] Backport account_cutoff_accrual_picking from 14 to 13: add prepaid support

alexis-via:
Also backport from 14 to 13 the field post_cutoff_move on res.company and the field source_move_state on account.cutoff


alexis-via:
This PR also fixes the broken layout on the accounting config page.

PR #231, opened 2023-02-15
[16.0][MIG] account_cutoff_accrual_subscription

alexis-via:
This migration uses this v14 PR as starting point https://github.com/OCA/account-closing/pull/224


alexis-via:
test error is unrelated

PR #230, opened 2023-02-15
[16.0][MIG] account_cutoff_picking (ex account-cutoff_accrual_picking)

alexis-via:
The migration has been made from this v14 PR that adds prepaid support https://github.com/OCA/account-closing/pull/225

PR #235, opened 2023-02-23
[16.0][MIG] ```account_cutoff_picking``` (ex ```account_cutoff_accrual_picking```)

yankinmax:
This PR is opened to supersede that one opened by @alexis-via:
- [ ] https://github.com/OCA/account-closing/pull/230

I just rewrote commit message as @simahawk suggested here:
https://github.com/OCA/account-closing/pull/230#issuecomment-1436687565


account_cutoff_picking/migrations/16.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not openupgrade.column_exists(
+        env.cr, "res_company", "default_cutoff_picking_interval_days"
+    ):
+        openupgrade.rename_fields(
```
simahawk, line 13: wondering: shouldn't we move fields to the new module name?

yankinmax: what do you mean?
this what is done in this migration script for `res.company`
I suppose it's needed only for this field, because fresh installation won't cause any problems.
And this is proposed to migrate from v15 to v16 on existing projects.

yankinmax: there are only two fields introduced in this module:
`default_cutoff_accrual_picking_interval_days` -> `default_cutoff_picking_interval_days`:needs to be renamed
`picking_interval_days` name stays the same
https://github.com/OCA/account-closing/blob/f0fde8029c384c72f50f002ecde43de090dce697/account_cutoff_accrual_picking/models/account_cutoff.py#L15

account_cutoff_picking/models/account_cutoff.py
```diff
@@ -148,14 +150,14 @@ def order_line_update_oline_dict(
             "precut_delivered_logs": [],
             "precut_invoiced_qty": 0.0,  # in product_uom
             "precut_invoiced_logs": [],
-            "name": _("%s: %s") % (order.name, order_line.name),
+            "name": " ".join([order.name, order_line.name]),
```
StephaneMangin: Missing a ':' inside the string

StephaneMangin (approved):

yankinmax:
> @simahawk your merge command was aborted due to failed check(s), which you can inspect on [this commit of 16.0-ocabot-merge-pr-235-by-simahawk-bump-nobump](https://github.com/OCA/account-closing/commits/d16d826ce1f077ad7500cfaee01cc978ed1f657c).
> 
> After fixing the problem, you can re-issue a merge command. Please refrain from merging manually as it will most probably make the target branch red.

this should be fixed now

PR #243, opened 2023-04-05
[16.0][ADD] account_invoice_start_end_dates_tax_change

sebalix:
Glue module between `account_invoice_start_end_dates` and
- https://github.com/OCA/account-invoicing/pull/1433

PR #244, opened 2023-04-07
[10.0][ADD] account_invoice_start_end_dates_tax_change

sebalix:
Glue module between `account_invoice_start_end_dates` and
- https://github.com/OCA/account-invoicing/pull/1435

These modules are backports of the following PRs:
- https://github.com/OCA/account-invoicing/pull/1433
- https://github.com/OCA/account-closing/pull/243


account_invoice_start_end_dates_tax_change/wizards/account_move_apply_tax_change.py
```diff
@@ -0,0 +1,74 @@
...
+    :return: the resulting date/datetime.
+    """
+    return value - relativedelta(*args, **kwargs)
+
+
+class AccountMoveApplyTaxChange(models.TransientModel):
+    _inherit = "account.move.apply.tax.change"
+
+    def _skip_line(self, line):
+        if line.start_date or line.end_date:
```
grindtildeath:
```suggestion
        if line.start_date and line.end_date:
```
Otherwise we'll probably get a TypeError comparing Date and Boolean types.

account_invoice_start_end_dates_tax_change/wizards/account_move_apply_tax_change.py
```diff
@@ -0,0 +1,74 @@
...
+            if line.start_date <= self.tax_change_id.date < line.end_date:
+                return False
+            return True
+        return super(AccountMoveApplyTaxChange, self)._skip_line(line)
+
+    def _change_taxes(self, line):
+        if line.start_date or line.end_date:
+            # Start & end dates defined in the scope of the tax changes
+            if line.start_date <= self.tax_change_id.date < line.end_date:
```
grindtildeath:
```suggestion
        if self._is_recurring_line():
            if self._is_recurring_tax_change_required():
(...)
    def _is_recurring_line(self, line):
        return line.start_date and line.end_date
    
    def _is_recurring_tax_change_required(self, line):
        # Start & end dates defined in the scope of the tax changes
        return line.start_date <= self.tax_change_id.date < line.end_date
```
I guess we can refactor to avoid using the same comparison in this function and the one above

account_invoice_start_end_dates_tax_change/wizards/account_move_apply_tax_change.py
```diff
@@ -0,0 +1,74 @@
...
+                tax_change_date = fields.Date.from_string(self.tax_change_id.date)
+                new_end_date = date_subtract(tax_change_date, days=1)
+                # Dates subtractions are not inclusives so we add +1 day each time
+                end_date = fields.Date.from_string(line.end_date)
+                start_date = fields.Date.from_string(line.start_date)
+                total_days = (end_date - start_date).days + 1
+                days_until_change = (new_end_date - start_date).days + 1
+                days_from_change = (end_date - tax_change_date).days + 1
+                price_unit1 = line.price_unit * days_until_change / total_days
+                price_unit2 = line.price_unit * days_from_change / total_days
```
grindtildeath:
```suggestion
                price_unit2 = line.price_unit - price_unit1
```
:warning: we have to make sure the sum of both new price units equals the original one, which could not be the case with precision roundings.

account_invoice_start_end_dates_tax_change/wizards/account_move_apply_tax_change.py
sebalix: I did it this way as the start/end dates module already make sure to have both date fields filled up if one of them is set. But yes, better with an `and` regarding the code below.

account_invoice_start_end_dates_tax_change/wizards/account_move_apply_tax_change.py
sebalix: done

PR #254, opened 2023-06-29
[16.0] account_cutoff_base: simplify config for taxes

alexis-via:
For all countries where the provision account for taxes is the same for all taxes, you can now configure the 2 provision accounts for taxes on the accounting configuration page.
This commit also fixes a bug in the generation of the move when you have a default tax on the revenue/expense account.


rlemeur (approved):

njeudy (approved):
Thanks alexis, tested and approved (and use on my database)

PR #255, opened 2023-07-06
[14.0][ADD] new module account_invoice_start_end_dates_move

sergiocorato:
This module enable to select start and end date for invoice lines directly on invoice header.


simona-buccheri (approved):
Testata la Pr funziona tutto correttamente
Simona

account_invoice_start_end_dates_move/readme/CONFIGURE.rst
```diff
@@ -0,0 +1 @@
+Nothing todo.
```
petrus-v: as far I understand users can choose to toogle the `apply_dates_all_lines`, it could be nice to write few lines about !

account_invoice_start_end_dates_move/models/account_move.py
```diff
@@ -0,0 +1,78 @@
...
+            lambda x: self.company_id.apply_dates_all_lines
+            or x.product_id.must_have_dates
+        ):
+            line.start_date = line.move_id.start_date
+            line.end_date = line.move_id.end_date
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "start_date" in vals:
+            lines = self.mapped("line_ids").filtered(
```
petrus-v: styling: `mapped` is useless here !

```suggestion
            lines = self.line_ids.filtered(
```

account_invoice_start_end_dates_move/models/account_move.py
```diff
@@ -0,0 +1,78 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    start_date = fields.Date(index=True)
+    end_date = fields.Date(index=True)
```
petrus-v: do you really needs to index those fields ?

account_invoice_start_end_dates_move/models/account_move.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "start_date" in vals:
+            lines = self.mapped("line_ids").filtered(
+                lambda l: l.start_date != vals["start_date"]
+                and self.company_id.apply_dates_all_lines
+                or l.product_id.must_have_dates
+            )
+            lines.write({"start_date": vals["start_date"]})
```
petrus-v: I'm wondering what's happen if:

* user open an existing invoice on edit mode
* user change the start date that will change the date on account move lines let's say `2023-07-25`
* user change the start date on the first line only to the `2023-07-26` and others lines kept to the `2023-07-25`
* user save the whole invoice

I'm not sure how odoo manage writes order between account.move and account.move.lines, does the fist line will be `2023-07-26` according this write ?

account_invoice_start_end_dates_move/tests/test_account_invoice_start_end_dates_move.py
```diff
@@ -0,0 +1,171 @@
...
+                    "product_id": self.insur_product.id,
+                    "name": "Insurance",
+                    "price_unit": 100,
+                    "quantity": 1,
+                    "account_id": self.account_revenue.id,
+                }
+            )
+        )
+        line._onchange_product_id()
+        line._convert_to_write(line._cache)
```
petrus-v, line 136: I'm wondering if you know odoo.tests.common.Form which is usefull to mimic onchange in unittest !?

petrus-v: hum ... actually you are not testing onchange here, probably the comment is wrong :thinking:

petrus-v (commented):
This module will be very usefull ! Thanks.

I've few questions and notices, also I would love a bit more coverage in trick write parts !

account_invoice_start_end_dates_move/readme/CONFIGURE.rst
sergiocorato: Done, thanks!

account_invoice_start_end_dates_move/models/account_move.py
sergiocorato: I used `mapped` because `self` is possibly a recordset, otherwise I should use a `for`.

account_invoice_start_end_dates_move/models/account_move.py
sergiocorato: Probably not, I've removed them, thanks!

account_invoice_start_end_dates_move/models/account_move.py
sergiocorato: Correct, this method would be confusing, I've removed it as the onchange is already doing the date synchronization.

account_invoice_start_end_dates_move/tests/test_account_invoice_start_end_dates_move.py
sergiocorato: I've given it a try, but this code fail as the onchanges are called at row creation, perhaps I'm missing something?
```
       with Form(
            self.env["account.move.line"]
            .with_user(self.account_user)
            .with_context(check_move_validity=False,
                          currency_id=invoice.currency_id)
        ) as form:
            form.move_id = invoice
            form.product_id = self.insur_product
            form.name = "Insurance"
            form.price_unit = 100
            form.quantity = 1
            form.account_id = self.account_revenue
            form.save()
```

account_invoice_start_end_dates_move/models/account_move.py
petrus-v: it shouldn't be a matter, ie:


```python
>>> env["account.move"].search([]).line_ids.move_id.mapped("name")
['INV/2023/07/0004', 'INV/2023/07/0003', 'INV/2023/07/0002', 'INV/2023/07/0001', 'BILL/2023/07/0001', 'BNK1/2023/01/0007', 'BNK1/2023/01/0006', 'BNK1/2023/01/0005', 'BNK1/2023/01/0004', 'BNK1/2023/01/0003', 'BNK1/2023/01/0002', 'BNK1/2023/01/0001', 'BILL/2018/09/0001']
```

account_invoice_start_end_dates_move/tests/test_account_invoice_start_end_dates_move.py
petrus-v: I'm using something like this:

```python

        move_form = Form(
            cls.env["account.move"].with_context(
                company_id=cls.fr_company.id, default_move_type="out_invoice"
            )
        )
        move_form.date = fields.Date.from_string("2022-06-15")
        move_form.invoice_date = fields.Date.from_string("2022-06-15")
        move_form.partner_id = cls.customer
        with move_form.invoice_line_ids.new() as line_form:
            line_form.product_id = cls.product

        cls.invoice = move_form.save()
```

petrus-v (approved):

PR #266, opened 2023-09-25
[16][ account_cutoff_picking] Avoid creating null cut-off line

florian-dacosta:
This may be the case of free products like samples for example.

@alexis-via


florian-dacosta:
@alexis-via

alexis-via (approved):

PR #268, opened 2023-10-23
[16.0][REF] account_cutoff_picking

jbaudoux:
Port and cleanup https://github.com/OCA/account-closing/pull/73

cc @alexis-via @florian-dacosta @lmignon @rousseldenis 

Instead of analyzing all pickings & invoices inside a given interval, only pickings and invoices generated after the cutoff date are analyzed, otherwise we rely on the SO/PO received/invoiced quantities. This allows to process a large amount of data in a reasonable amount of time. To decrease the amount of pickings & invoices to analyze a cron pre-generates the cutoff entries at the end of the period. Nevertheless, it is always possible to regenerate entries at any time (but the more pickings & invoices to analyze are, the more time it will take).
The limit in time in the past is then now disabled.

You can configure to disable accrual on closed orders.


jbaudoux:
Being replaced by https://github.com/OCA/account-closing/pull/271

PR #271, opened 2023-10-25
[16.0][ADD] account cutoff accrual on order

jbaudoux:
Port and refactor https://github.com/OCA/account-closing/pull/73

Generate cutoff accruals by analyzing order lines invoiced & delivered quantity (or ordered quantity for sales with invoice policy on order).

Several modules are provided:
* `account_cutoff_accrual_order_base`: Base module to analyze any type of order lines
* `account_cutoff_accrual_order_stock_base`: Base module to analyze order lines delivered by stock moves
* `account_cutoff_accrual_sale`: analyze sales order lines to accrue what must be invoiced. Generates accrued revenue, i.e. invoice to establish
* `account_cutoff_accrual_purchase`: analyze purchase order lines to accrue what must be invoiced. Generates accrued expense, i.e. invoice to receive
* `account_cutoff_accrual_sale_stock`: analyze sales order lines to accrue what must be delivered. Generates accrued prepaid revenue, i.e. goods to deliver
* `account_cutoff_accrual_purchase_stock`: analyze purchase order lines to accrue what must be received. Generates accrued prepaid expense, i.e. goods to receive

The modules allow to process a large amount of data in a reasonable amount of time. To decrease the amount of stock moves & invoices to analyze a cron pre-generates the cutoff entries at the end of the period. Nevertheless, it is always possible to regenerate entries at any time (but the more stock moves & invoices to analyze are, the more time it will take). Any invoice posted after the cutoff is generated will alter the corresponding cut-off entries. Once the cut-off is posted is accounting, you cannot anymore alter the entries.

Note: Those modules covers the feature provided by `stock_cutoff_picking` but in a more generic and efficient way. Instead of analyzing all pickings & invoices inside a given interval, only pickings and invoices generated after the cutoff date are analyzed, 
The limit in time in the past is then now disabled.

Accrual on order flagged as force_invoiced is disabled.

Road-map: The generic approach of the modules allows to extend the features to generate accrual on service SO / PO with project tasks

cc @alexis-via @florian-dacosta @rousseldenis


jbaudoux:
@alexis-via Added dependencies on sale/purchase_force_invoiced to handle exclusion. I dropped my boolean in favor of that `force_invoiced`

jbaudoux:
@alexis-via Everything we discussed as been done (that I remember as you forgot to post your review). Can you redo a final review?

jbaudoux:
Fixed cron job method

account_cutoff_accrual_sale/models/sale_order_line.py
```diff
@@ -25,13 +25,41 @@ def _get_cutoff_accrual_partner(self):
...
+            "product_tmpl_id",
+            self.env["product.template"]._table,
+            "id",
+            "product_tmpl_id",
+        )
         query.add_where(
-            f'"{self._table}".display_type IS NULL AND '
-            f'"{self._table}".qty_delivered != "{self._table}".qty_invoiced'
+            f"""
+            "{self._table}".display_type IS NULL AND
```
rousseldenis: Usage of AsIs() should be recommended using self._table

account_cutoff_accrual_sale/models/sale_order_line.py
```diff
@@ -0,0 +1,116 @@
...
+
+    is_cutoff_accrual_excluded = fields.Boolean(
+        compute="_compute_is_cutoff_accrual_excluded",
+        store=True,
+    )
+
+    @api.depends("order_id.force_invoiced")
+    def _compute_is_cutoff_accrual_excluded(self):
+        for rec in self:
+            rec.is_cutoff_accrual_excluded = rec.order_id.force_invoiced
```
rousseldenis, line 31: Why not only 'invoice_status' as force_invoiced will change it ?

rousseldenis (approved):
Code review. Some questions

rousseldenis:
@alexis-via This deserves a little review. Thanks

account_cutoff_accrual_sale/models/sale_order_line.py
jbaudoux: I trigger cutoff update on manual action on force invoiced flag. I don't want to trigger it each time a sale invoice status changes. And I don't know a good way to trigger some code execution on change of a stored computed field.

rousseldenis: > I trigger cutoff update on manual action on force invoiced flag. I don't want to trigger it each time a sale invoice status changes. And I don't know a good way to trigger some code execution on change of a stored computed field.

So, cutoff trigger does not occur at each time a SO line quantity is delivered and invoiced (can occur at several different times) ?

jbaudoux: * not at delivery as you cannot deliver in the past
* not at invoice creation (SO line qty_invoiced) but at invoice posting as this is when it is accounted

This makes me think that I cannot rely at all on the invoice status as this is changed at invoice creation and not when posted. So, I'll have to revert this change in the query domain.

jbaudoux: No, it's all good, I process also the draft invoices at https://github.com/OCA/account-closing/pull/271/files#diff-df279c387c73fffb2fef47c05b5c5dc189e1ddb956a175de21a5a4f69300af55R64

jbaudoux:
I fixed the case of sale delivery line (service) added after SO confirmation and after cutoff date.

jbaudoux:
I added the use case where a SO having only the delivery line (shipping cost) to invoice among the SO lines is not considered as invoiceable.

jbaudoux:
@alexis-via Have you had the chance to test this PR in comparison to your previous implementation ? Can it get merged ?

adrienpeiffer (approved):

jbaudoux:
@adrienpeiffer I added the missing readmes and squashed the fix commits

PR #272, opened 2023-10-25
[IMP][14.0] account_fiscal_year_closing do not split by partner account of type not in receivable or payable

sergiocorato:
The current method create an account.move.line for every partner found in every account.account, which is unuseful (account like assets or bank or every other different of a credit or debit).

With this fix, only the account of type 'payable' or 'receivable' are splitted by partner.


GSLabIt (approved):

francesco-ooops:
@sergiocorato should this be merged?

sergiocorato:
> @sergiocorato should this be merged?

This one is useful in my opinion.

PR #274, opened 2023-10-26
[16.0][FIX] account_cutoff_base tax display in view

jbaudoux:
Fix tax not displayed in lines full screen view


jbaudoux:
@alexis-via After our discusson, I dropped the company removal from the domain to keep only the fix on the tax display

alexis-via (approved):

PR #288, opened 2024-01-08
[FIX] account_fiscal_year_closing move line created with correct currency

sergiocorato:


account_fiscal_year_closing/models/account_fiscalyear_closing.py
```diff
@@ -604,6 +604,15 @@ def move_line_prepare(self, account, account_lines, partner_id=False):
...
                     "partner_id": partner_id,
                 }
+                if (
+                    account.currency_id
+                    and account.currency_id
+                    != self.fyc_config_id.fyc_id.company_id.currency_id
+                ):
+                    move_line.update(
+                        currency_id=account.currency_id.id,
+                        amount_currency=balance,
```
lk-eska: balance calculation is incorrect. this is the balance in company currency. we get it working like this:
```

                if account.currency_id and account.currency_id != account.company_currency_id:
                    move_line.update(
                        currency_id=account.currency_id.id,
                        amount_currency=sum(account_lines.mapped('amount_currency')) * -1,
                    )

```

sergiocorato: Thanks @lk-eska , I've updated the fix, anyway `account.company_currency_id` does not exist afaik.

sergiocorato: <strike>I computed the value of amount_currency on the balance of the account, with the date of the closing move, as the direct sum of the currency amounts is not meaningful as they are referred to many dates.</strike>

PR #289, opened 2024-01-15
[IMP] account_cutoff_base: Add index on cutoff line to move line

rousseldenis:
In case of account move line deletion, this can speedup about 500ms per line.


sbidoul (approved):
Important when you need to delete a move for any reason.

jbaudoux (approved):

Issue #280, opened 2023-11-09
Bug in Fiscal Year Closing module in odoo version 14


Evaguadaltech:
![error-cierre-ejercicio-comufri](https://github.com/OCA/account-closing/assets/122524105/d378bd94-f372-4da4-8904-334a16841e7d)
Hello, when we have created a year-end closing and we have added the transaction configuration. When saving or calculating the year-end closing it gives the attached error. And it does not create the year-end closing.

Thank you.

DesarrolloWIL:
I have the same problem, I am investigating the origin of it.

Evaguadaltech:
Buenos días, ¿se ha podido corregir a dia de hoy la incidencia que he indicado en odoo 14?

Gracias.

mide-dynapps:
Hi, same issue.
I thinks it's linked to PR https://github.com/OCA/account-closing/pull/199 for the migration into v14

Any chances to revert commit https://github.com/OCA/account-closing/commit/fb4c3ee760231b066f4085984e8abcbf8f4b2b56 ?
I don't understand the need for this code :)
And w/o this code, it seems to work fine.

mide-dynapps:
**How to reproduce:**
- settings => for current user, set checkbox on user profile 'Show Accounting Features - Readonly'
![image](https://github.com/OCA/account-closing/assets/79934758/cfedb9a8-63a3-421b-9b9b-f88086152da4)

- Invoicing / Accounting / Fiscal year closing => create new entry:
![image](https://github.com/OCA/account-closing/assets/79934758/29a8bef8-dd73-4f2d-b2bd-32591195ee8a)
![image](https://github.com/OCA/account-closing/assets/79934758/2d99fff1-8e43-4348-a047-0e9ca80d1c0c)

**Result:** Impossible to Save or Calculate => crash with traceback

Evaguadaltech:
Hola, el error lo daba cuando el campo Destination account lo pongo vacio cuando el campo Move type es Closing. ¿me habeis entendido?

Gracias.
Un saludo.

mide-dynapps:
@Evaguadaltech any chance to test the PR #294 ?

Evaguadaltech:
Hola, los cambios que has indicado los he probado en local y me han funcionando correctamente. 

Gracias. 
Un saludo.

PR #248, opened 2023-06-14
[16.0] [MIG] account_fiscal_year_closing

Borruso:
Migration account_fiscal_year_closing from 14.0 to 16.0

--
Confermo di aver firmato il CLA https://odoo-community.org/page/cla e di aver letto le linee guida su https://odoo-community.org/page/contributing

PR #294, opened 2024-03-14
[14.0] [FIX] account_fiscal_year_closing - create/write

mide-dynapps:
When creating new fiscal year closing with an account mapping (with destination account) in moves configuration, an issue is raised at save (Issue https://github.com/OCA/account-closing/issues/280).

But when coming from a template (selecting template on screen), Mapping data are used as an Array instead instead of simple id (due to many inside another many). 

The current code manage only case from template.

This PR take into account this particularity (and mangage other way), check the type before taking the id of the array.

### DIFF
--- a/account_fiscal_year_closing/models/account_fiscalyear_closing.py
+++ b/account_fiscal_year_closing/models/account_fiscalyear_closing.py
@@ -553,13 +553,13 @@ class AccountFiscalyearClosingMapping(models.Model):
 
     @api.model
     def create(self, vals):
-        if "dest_account_id" in vals:
+        if "dest_account_id" in vals and type(vals["dest_account_id"]) == list:
             vals["dest_account_id"] = vals["dest_account_id"][0]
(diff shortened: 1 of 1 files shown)


mide-dynapps:
Bug also on v16.0 #248

lk-eska:
I confirm the bug and fix. Please merge this.

account_fiscal_year_closing/models/account_fiscalyear_closing.py
```diff
@@ -553,13 +553,13 @@ class AccountFiscalyearClosingMapping(models.Model):
 
     @api.model
     def create(self, vals):
-        if "dest_account_id" in vals:
+        if "dest_account_id" in vals and type(vals["dest_account_id"]) == list:
```
Borruso, line 556: use isinstance(vals["dest_account_id"], list)

Borruso (changes requested):
Thanks for PR

lk-eska:
Would you please merge this?

PR #300, opened 2024-07-09
[LIC] account_invoice_start_end_dates: change to LGPL

sbidoul:
I propose to change the license of account_invoice_start_end_dates to LGPL for interoperability with enterprise addons.

Since this modules provides fields with little business logic, as foundation for higher level modules, this typically enter the category of modules that are good candidates for LGPL.

@alexis-via as author, do you agree?


sbidoul:
@alexis-via gentle nudge here: would you agree to this license change? We would then request the same for sale_start_end_date.

alexis-via (approved):
I'm ok to change the licence to LGPL for this small module

rvalyi:
> I'm ok to change the licence to LGPL for this small module

great move @alexis-via !  Let's keep adding great AGPL modules but that's cool we be more permissive on these very simple modules meanwhile.

rvalyi (changes requested):
@sbidoul you forgot to change the file headers. I did it in this little PR for convenience https://github.com/acsone/account-closing/pull/5

After that it will be OK for me, nice move from @alexis-via

sbidoul:
@rvalyi good catch, thanks! I merged your PR and squashed.

rvalyi (approved):

sbidoul:
FW port to 17 in #333

sbidoul:
And FW port to 18 in #334

PR #304, opened 2024-09-22
[14.0][FIX] account_move_cutoff: while using full refund wizard

petrus-v:
action_post is not called better to overload _post on account.move model


barkat-matthias (approved):
LGTM

damdam-s (approved):

PR #307, opened 2024-10-30
18.0 mig account invoice start end dates

fredericgrall:

PR #309, opened 2024-11-14
[18.0][MIG] account_invoice_start_end_dates: Migration to 18.0

xaviedoanhduy:
supersedes: https://github.com/OCA/account-closing/pull/307


account_invoice_start_end_dates/tests/test_invoice_start_end_dates.py
```diff
@@ -0,0 +1,94 @@
...
+                            "account_id": self.account_revenue.id,
+                            "start_date": self._date("01-01"),
+                            "end_date": self._date("06-30"),
+                        },
+                    ),
+                    (
+                        0,
+                        0,
+                        {
+                            "product_id": self.env.ref("product.product_product_5").id,
```
ajaniszewska-dev, line 84: looking at this test case i guess we test positive case here - all previous lines have dates set and we don't expect any error to be raised for product_product_5 because probably this product has no flag set - i'd maybe add some comments in this case + test negative case (where error should be raised)?

ajaniszewska-dev (approved):
LG, i'd slightly improve test

account_invoice_start_end_dates/tests/test_invoice_start_end_dates.py
xaviedoanhduy:
```
if moveline.start_date and not moveline.end_date:
    ...
if moveline.end_date and not moveline.start_date:
    ...
if (
    moveline.end_date
    and moveline.start_date
    and moveline.start_date > moveline.end_date
):
```
hi @ajaniszewska-dev, I had the same question as you, I'm not sure what the author of this module intended, because the above conditions require at least 1 value to be assigned to `start_date` and `end_date`, so if no value is inputed, no error will be raised. I tried adding a filter in the lookup function like:
```
def _check_start_end_dates(self):
    for moveline in self.filtered("must_have_date"):
         ...
```
then the error will be raised, and I wonder if that change affects the original purpose of this module?

ajaniszewska-dev: @alexis-via  probably your input here would be valuable.

alexis-via:
I tried your PR, but faced several issues:

mig_closing_popup

and also an issue in product form view (at least when using with intrastat_base... it's kind of a decision if we should create groups inside the native group name="accounting" (choice made by intrastat_base) or put fields directly in it.

closing_product_form_view

So, for these 2 reasons, I decided to make another PR https://github.com/OCA/account-closing/pull/313

account_invoice_start_end_dates/tests/test_invoice_start_end_dates.py
alexis-via: I added the negative tests in my migration PR https://github.com/OCA/account-closing/pull/313

PR #310, opened 2024-11-14
[IMP] Move the constraint on uniqueness for cutoff date at line level…

lmarion-source:
… + Add the possibility to create several cutoff with the same date

Business case can be :

- One user (A) creates a cutoff for some account moves. Another user (B) creates other moves, then wants to see those moves in a cutoff. However, B does not want to interfere with the cutoff A already made. B creates a new cutoff with the same date as the one of A, and when he generates lines, only retrieves the one that were not already in the cutoff of A.

- One user wants to create several cutoff at the same date, but for different journals/accounts

### DIFF
--- a/account_cutoff_base/__manifest__.py
+++ b/account_cutoff_base/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "Account Cut-off Base",
-    "version": "16.0.1.5.0",
+    "version": "16.0.1.5.1",
     "category": "Accounting & Finance",
     "license": "AGPL-3",
     "summary": "Base module for Account Cut-offs",
--- a/account_cutoff_base/migrations/16.0.1.5.1/post-migrate.py
+++ b/account_cutoff_base/migrations/16.0.1.5.1/post-migrate.py
@@ -1,0 +1,19 @@
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
+
+import odoo
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    _logger.info("Remove constraint on account cutoff")
+    odoo.api.Environment(cr, odoo.SUPERUSER_ID, {})
+
+    cr.execute(
+        """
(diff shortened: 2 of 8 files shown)


account_cutoff_base/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "Account Cut-off Base",
-    "version": "16.0.1.5.0",
+    "version": "16.0.1.0.1",
```
jbaudoux: You cannot go backward

jbaudoux (changes requested):
In my opinion, you cannot do such change because it has an impact on many other cutoff modules. Can't you leverage the user on cutoff entries instead of creating a separate new cutoff?

Side note: you should make one commit per module

account_cutoff_base/migrations/16.0.1.0.1/post-migrate.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def migrate(cr, version):
+    _logger.info("Remove constraint on account cutoff")
+    odoo.api.Environment(cr, odoo.SUPERUSER_ID, {})
+
+    cr.execute(
+        """
+        ALTER TABLE account_cutoff
+        DROP CONSTRAINT account_cutoff_date_type_company_uniq
```
acsonefho: `IF EXISTS`

acsonefho (changes requested):

acsonefho (approved):

acsonefho:
@OCA/accounting-maintainers Is it possible to re-open this one and review/merge?
It's already used in production for months :pray:

jbaudoux:
> @OCA/accounting-maintainers Is it possible to re-open this one and review/merge? It's already used in production for months 🙏

But I'm against this change. And there is no reply/follow-up to my comment

rousseldenis:
@lmarion-source @acsonefho @sbidoul Could you attend comments made by @jbaudoux ?

lmarion-source:
> @lmarion-source @acsonefho @sbidoul Could you attend comments made by @jbaudoux ?

On it :)

PR #311, opened 2024-12-06
[18.0] [MIG] account_cutoff_base: Migration to 18.0

Borruso:
Migration from 17.0 to 18.0

Issue: https://github.com/OCA/account-closing/issues/306


account_cutoff_base/models/account_cutoff.py
```diff
@@ -118,7 +118,7 @@ def _default_cutoff_account_id(self):
     cutoff_account_id = fields.Many2one(
         comodel_name="account.account",
         string="Cut-off Account",
-        domain="[('deprecated', '=', False), ('company_id', '=', company_id)]",
+        domain="[('deprecated', '=', False)]",
```
alexis-via, line 121: You should keep company in the domain and replace it with ('company_ids', 'in', company_id)
This problem is present at several places in this commit.
So I propose another migration PR: https://github.com/OCA/account-closing/pull/329

PR #313, opened 2024-12-18
[18.0][MIG] account_invoice_start_end_dates

alexis-via:
Migration + some code and test improvements (in a separate commit)


alexis-via:
I also added the fields start_date and end_date in account.invoice.report.

PR #326, opened 2025-02-12
[16.0][FIX] account_cutoff_accrual_sale : Fixes for SO invoiced on order + [ADD] support for down payment

jbaudoux:
I added pleeeeenty of tests !

Covering:
* normal service on order : should be invoiced when SO is confirmed
* shipping service on order : should be invoiced when first stock delivery occurs (that's a specificity of delivery module https://github.com/odoo/odoo/blob/16.0/addons/delivery/models/sale_order.py#L155)
* stock product on order : should be invoiced when delivered (invoicing policy doesn't matter here, accrual is based on stock valuation change, i.e. the effective stock delivery)

cc @rousseldenis @lmignon @sbidoul 

Also fixed `account_cutoff_accrual_order_base` that was updating stock in the test while not depending on `stock` module. Code moved to modules depending on this one and `stock`

**Edit:** I also added support for downpayment


account_cutoff_accrual_sale/models/sale_order_line.py
```diff
@@ -106,6 +157,14 @@ def _get_cutoff_accrual_delivered_stock_quantity(self, cutoff):
...
+        # In case of stock, we always consider what is delivered as this
+        # impacted the stock valuation.
         return self.qty_delivered
+
+    def _get_cutoff_accrual_delivered_min_date(self):
+        """Return first delivery date"""
+        self.ensure_one()
+        if self.product_id.invoice_policy == "order":
+            date = self.order_id.date_order
+            return date.date()
```
jbaudoux: Need to use right TZ

jbaudoux:
@lmignon @rousseldenis This needs review

rousseldenis (approved):
LGTM

jbaudoux:
@alexis-via @sbidoul Can you merge this one?

jbaudoux:
@sbidoul and merge this one also

PR #341, opened 2025-04-11
[16.0][FIX] account_cutoff_accrual_order_base: product analytic

jbaudoux:
Keep product on each move line.
Set the move type to allow product_analytic to compute the right analytic account.

Depends on:
* https://github.com/OCA/account-analytic/pull/768

~~This change will require to modify the accrual journal to make it a sales or purchase journal if it was a miscellaneous journal. This is required to map to the right income or expense analytic account.~~

cc @sbidoul


rousseldenis (approved):

lmignon (approved):

sbidoul:
> This change will require to modify the accrual journal to make it a sales or purchase journal if it was a miscellaneous journal

This is surprising and, I think, problematic. Why is this necessary?

jbaudoux:
> > This change will require to modify the accrual journal to make it a sales or purchase journal if it was a miscellaneous journal
> 
> This is surprising and, I think, problematic. Why is this necessary?

@sbidoul Yes, that's why I wanted your feedback. This is due to the fact that there is an income and expense analytic account on the product and to apply the right one, the account move must belong to an sales or purchase journal. See https://github.com/OCA/account-analytic/blob/16.0/product_analytic/models/account_move.py#L46
I don't see that as an issue. Do you?
In odoo standard, when you accrue from the PO, it's making an entry on a "entry" journal. So won't work neither.
Do you see another approach to make it compatible with "entry" journals ?

sbidoul:
I'm afraid having accrual moves on sale/purchase journals will create problems down the line. For instance in tax reporting it's likely that the journal type is involved.

When we do accruals from invoices, we use the same analytic distribution as the original moves. So maybe we should compute the analytic distribution in the accrual move creation code, using the same logic that computes it when creating an invoice from SO/PO?

jbaudoux:
@sbidoul I modified `product_analytic` to allow to collect the right distribution without having to change the journal. https://github.com/OCA/account-analytic/pull/768
I'll adapt this PR

jbaudoux:
TODO: add test

jbaudoux:
> TODO: add test

done

account_cutoff_accrual_purchase_product_analytic/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_cutoff_expense
```
rousseldenis, line 1: @jbaudoux Problem in file naming here

PR #343, opened 2025-06-12
[IMP] Move the constraint on uniqueness for cutoff date at line level using a configuration

lmarion-source:
+ Add the possibility to create several cutoff with the same date

If the box is not checked in configurations,  the uniqueness is checked at the cutoff level (as previously done). If we need to create several cutoff at the same date for different journals/accounts, we can check the box and the check will be done at the cutoff line level.

Business case can be :

- One user (A) creates a cutoff for some account moves. Another user (B) creates other moves, then wants to see those moves in a cutoff.
  However, B does not want to interfere with the cutoff A already made. B creates a new cutoff with the same date as the one of A, and when he generates lines,
  only retrieves the one that were not already in the cutoff of A.

- One user wants to create several cutoff at the same date, but for different journals/accounts

### DIFF
--- a/account_cutoff_base/migrations/16.0.1.5.1/post-migrate.py
+++ b/account_cutoff_base/migrations/16.0.1.5.1/post-migrate.py
@@ -1,0 +1,19 @@
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
+
+import odoo
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    _logger.info("Remove constraint on account cutoff")
+    odoo.api.Environment(cr, odoo.SUPERUSER_ID, {})
+
+    cr.execute(
+        """
+        ALTER TABLE account_cutoff
+        DROP CONSTRAINT IF EXISTS account_cutoff_date_type_company_uniq
+    """
+    )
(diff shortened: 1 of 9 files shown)


lmarion-source:
@jbaudoux I was not able to re-open the #310 so I did a new one and add a configuration for the business case I have. This way the original behaviour stays the same

acsonefho (approved):
Code LGTM

account_cutoff_base/migrations/16.0.1.5.1/post-migrate.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
+
+import odoo
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
```
jbaudoux, line 10: there is no version change in the manifest for this script

account_cutoff_base/models/account_cutoff_line.py
```diff
@@ -83,3 +86,54 @@ class AccountCutoffLine(models.Model):
...
+
+            if any(r["__count"] > 1 for r in read_group):
+                raise UserError(
+                    _(
+                        "A cutoff line of the same type already exists with this cut-off date !"
+                    )
+                )
+        else:
+            read_group = self.env["account.cutoff"].read_group(
+                domain=[("id", "in", self.mapped("origin_move_line_id").ids)],
```
jbaudoux: there was no check on origin_move_line_id before. You don't limit to self here like in the other if?

jbaudoux (changes requested):
you need to make one commit per module

PR #351, opened 2025-10-08
[18.0][IMP] account_invoice_start_end_dates: enforce start_date+end_date only on posted invoices

alexis-via:
We enforce start_end+end_date when product_id.must_have_dates=True only when posting the invoice, because some users want to use the module account_invoice_start_end_dates WITHOUT the module sale_start_end_dates for a good reason.

This is a sort of revert of the change I made in dec 2024 in this commit: https://github.com/OCA/account-closing/commit/c988c2647f203864cdf10f0bf600dd25c020dc30


njeudy (approved):
Juste taste it, and work as wanted ! thanks.

Alexis-SDK (approved):
I approve

PR #283, opened 2023-11-17
[16.0][MIG] account_cutoff_base: Allow to configure different account for refunds (credit notes)

jbaudoux:
Forward port commit https://github.com/OCA/account-closing/commit/83e9b1c7f9463043414d88374f67c9ee01873b9a

PR #270, opened 2023-10-25
[16.0][IMP] account_cutoff_base: auto reverse

jbaudoux:
Allow to automatically reverse generated moves

Added test move creation

cc @alexis-via @rousseldenis

PR #357, opened 2026-01-19
[19.0][MIG] account_cutoff_base

alexis-via:


account_cutoff_base/models/account_cutoff_line.py
```diff
@@ -0,0 +1,82 @@
...
+from odoo import fields, models
+
+
+class AccountCutoffLine(models.Model):
+    _name = "account.cutoff.line"
+    _inherit = "analytic.mixin"
+    _check_company_auto = True
+    _description = "Account Cut-off Line"
+
+    parent_id = fields.Many2one("account.cutoff", string="Cut-off", ondelete="cascade")
```
petrus-v, line 14: I would have called it account_cutoff_id, parent/child are mainly used when model is pointing to itself

account_cutoff_base/models/account_cutoff.py
```diff
@@ -0,0 +1,461 @@
...
+        value = ID of cutoff_account"""
+        self.ensure_one()
+        mappings = self.env["account.cutoff.mapping"].search(
+            Domain(
+                [
+                    ("company_id", "=", self.company_id.id),
+                    ("cutoff_type", "in", ("all", self.cutoff_type)),
+                ]
+            )
+        )
```
petrus-v, line 404: I can see this method is used in account_cutoff_start_and_end dates module but I would rather prepare the mapping record with concerned record only so this method could use an optionnal source accounts.




```suggestion
    def _get_mapping_dict(self, source_accounts=None):
        """return a dict with:
        key = ID of account,
        value = ID of cutoff_account"""
        self.ensure_one()
        domain = Domain(
            [
                ("company_id", "=", self.company_id.id),
                ("cutoff_type", "in", ("all", self.cutoff_type)),
            ]
        )
        if source_accounts:
            domain &= Domain("account_id", "in", source_accounts.ids)
        mappings = self.env["account.cutoff.mapping"].search(
        )
```

petrus-v (approved):
I'm approving with some improvement suggestions. Also I would love to see more tests in this module, as an independent module it should cover its code.

account_cutoff_base/views/account_cutoff.xml
```diff
@@ -0,0 +1,224 @@
...
+                                name="source_move_state"
+                                widget="radio"
+                                readonly="state == 'done'"
+                            />
+                            <field
+                                name="company_id"
+                                options="{'no_create': True}"
+                                readonly="state == 'done'"
+                            />
+                            <field name="company_currency_id" invisible="1" />
```
AnizR, line 81: Can be safely removed since https://github.com/odoo/odoo/pull/137031

```suggestion
```

account_cutoff_base/views/account_cutoff.xml
```diff
@@ -0,0 +1,224 @@
...
+            <list decoration-info="state == 'draft'">
+                <field
+                    name="cutoff_type"
+                    readonly="state == 'done'"
+                    column_invisible="context.get('default_cutoff_type')"
+                />
+                <field name="cutoff_date" decoration-bf="1" />
+                <field name="move_ref" optional="hide" />
+                <field name="total_cutoff_amount" />
+                <field name="company_currency_id" column_invisible="1" />
```
AnizR, line 128: Can be safely removed

```suggestion
```

AnizR (approved):
Thanks for your contribution!

Code LGTM, just some nit picking.

petrus-v:
@alexis-via I've propose a PR on top of yours to take care of main reviews here.

the last but not least is the field renaming (parent_id to cutoff_id and cutoff_line_id), I've a commit that do it, it is working but will impact other modules, so do you think we can do it in this version or do you prefer to do it in the next version ?

petrus-v:
Could you consider integrate https://github.com/OCA/account-closing/pull/283 / https://github.com/OCA/account-closing/pull/270 in this PR ?

petrus-v:
I've create a new PR for this adding the previous proposal and forwarding v16 PRs => https://github.com/OCA/account-closing/pull/381

PR #365, opened 2026-02-13
[18.0][FIX] account_invoice_start_end_dates: Avoid same-label warning

tobiaszehntner:
Odoo's account_accountant has two fields with the same label on the same model. Avoid warning by renaming the fields in Python while keeping the same label in the views. 

Fields: `account_accountant/models/account_move.py:446`

Warnings:
```
WARNING odoo.addons.base.models.ir_model: Two fields (start_date, deferred_start_date) of account.move.line() have the same label: Start Date. [Modules: account_invoice_start_end_dates and account_accountant] 
WARNING odoo.addons.base.models.ir_model: Two fields (end_date, deferred_end_date) of account.move.line() have the same label: End Date. [Modules: account_invoice_start_end_dates and account_accountant] 
```

### DIFF
--- a/account_invoice_start_end_dates/models/account_move_line.py
+++ b/account_invoice_start_end_dates/models/account_move_line.py
@@ -10,8 +10,8 @@ from odoo.tools.misc import format_date
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
-    start_date = fields.Date(index=True)
-    end_date = fields.Date(index=True)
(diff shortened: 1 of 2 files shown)


acsonefho (approved):

tobiaszehntner:
@alexis-via could this be merged?

alexis-via:
Seems like an Odoo-EE only problem. But ok, let's do it.

PR #367, opened 2026-03-17
[FIX] account_invoice_start_end_dates: avoid check on deferred lines

tobiaszehntner:
The constrains raises for [generated deferred entries](https://github.com/odoo/enterprise/blob/4565d62391f939b5a1fb6702041e3d9e58b9f496/account_accountant/models/account_move.py#L114) even though the posted entry has valid dates.


tobiaszehntner:
Hello @alexis-via 
Could you confirm if this is the best way to narrow the constrains? Our issue is with Odoo Enterprise (account_accountant) but I wouldn't be able to check on those specific fields inside the OCA module.
Thanks, Tobias

tobiaszehntner:
@alexis-via could this be merged so it's not lost?

PR #382, opened 2026-07-16
[19.0] account_cutoff_base: Move the constraint on uniqueness for cutoff date at line level

petrus-v:
port #343 account_cutoff_base part on top of #381


account_cutoff_base/migrations/19.0.1.0.1/post-migrate.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 ACSONE SA/NV
```
petrus-v, line 1: I've to move this to 19.0.1.0.1 as far this script is idempotent and the formal pr was never merged in v16 seems secure

AnizR (approved):
I reviewed only the last commit that integrates

Code LGTM and I think that you fixed a bug here (using id of lines to search on parent): https://github.com/OCA/account-closing/pull/343/changes#diff-191f90fce2c30a95de8f7309b0045c5a9827577f933e93e9d4576cb5ba2d530dR128

Thanks!

PR #384, opened 2026-07-16
[19.0] account_cutoff_start_end_dates: Move the constraint on uniqueness for cutoff date at line level

petrus-v:
port #343 account_cutoff_start_end_dates part.

based on #382 depends on #382


account_cutoff_start_end_dates/models/account_cutoff.py
```diff
@@ -0,0 +1,254 @@
...
+        }
+        for rec in self:
+            source_journal_ids = self.env["account.journal"]
+            if rec.cutoff_type in mapping:
+                source_journal_ids = self.env["account.journal"].search(
+                    [
+                        ("type", "=", mapping[rec.cutoff_type]),
+                        ("company_id", "=", rec.company_id.id),
+                    ]
+                )
```
AnizR: I think that we could improve this from a performance point of view.
What about doing 1 search (or maybe a readgroup if we have multiple company) before the loop  then assigning your source journals?

(code should be modified and not taken as it is)
```
self.env["account.journal"].search(
                    [
                        ("type", "in", ["sale","purchase"]),
                        ("company_id", "in", self.company_id.ids),
                    ]
                )
```

petrus-v: Thanks for suggestion :)

done using read group !

AnizR (approved):
Code LGTM 👍

